PR #72, opened 2019-10-15
added select_best method

fahadshery:
this is a quick method i put together to get the best scores + params after tuning. Might need to amend later to include multiple `metrics` i.e. `select_best(metric = c("rmse","rsq"))`.


R/select_best.R
```diff
@@ -0,0 +1,75 @@
...
+        dplyr::select(mean) %>% dplyr::rename(!!paste0("best_",metric) := mean) %>%
+        dplyr::slice(1:n_top)
+    }
+    else if (metric %in% valid_desc_metrics & nrow(summary_res %>% dplyr::filter(.metric == metric)) !=0) {
+      summary_res %>%
+        dplyr::arrange(desc(mean)) %>%
+        dplyr::select(mean) %>% dplyr::rename(!!paste0("best_",metric) := mean) %>%
+        dplyr::slice(1:n_top)
+    } else stop("No results are available. Please check trained performance metrics returned by the model tunning")
+  }
```
topepo: It would be better to assign the returned value to an object (I usually use `res`) and return that at the end. It's a little bit more readable and, if more needs to be done with the object later, less invasive to change.

R/select_best.R
```diff
@@ -0,0 +1,75 @@
...
+select_best <- function(x, metric=NA,n_top = 1,performance = TRUE,maximize = NULL) {
+
+  if (is.na(metric)) {
+    stop("Please specify a character value for metric to get the best score/params...")
+  }
+
+  metric = tolower(metric)
+
+  #metrics that require ascending/desc sorting
+  valid_asc_metrics <- c("rmse" , "mase" , "rpd")
```
topepo: I would avoid hard-coding the metrics. We'll have a method (soon-ish) that will determine the direction. Most metrics should be maximized so go ahead and default it to that and remove lines 33 - 42. 

You should also check to see if it is a single character storing and use `rlang::abort()` if it does not (I'm trying to get away from `stop()` when I remember to do so).

topepo:
It looks good (esp for a first PR!). 

Minor comments in-line. 

Overall comments:

 * Run the code through a styler. You can highlight it in the RStudio IDE and use shift-command-A or use the styler add-in at the top of the IDE. That is far from perfect so give the [style guide](https://style.tidyverse.org/syntax.html) a quick read. 

 * Write some test cases. It's not terrible hard and, if you've never done that, check out that section in the [R packages book](http://r-pkgs.had.co.nz/tests.html). Overall, there should be test cases for every `if` statement and for things that you envision would be tricky (e.g. tied metrics in this case or NA values). There are examples in the package to use as your guide. 

 * Make sure to run `R CMD check` once you've made the changes to evaluate if the documentation and other aspects are correct.

R/select_best.R
fahadshery: there are few matrics that need ascending sorting to get the best result out. i.e. I want to see the lowest `rmse` out of all. However, if I want to see `rsq` then it will have to be in the `desc` order? If this doesn't matter then I will delete this. Please advice

R/select_best.R
fahadshery: done

R/select_best.R
```diff
@@ -0,0 +1,75 @@
...
+#' select_best(initial_grid_knn,metric = "rmse", performance = FALSE)
+#' }
+#' @export
+select_best <- function(x, metric=NA,n_top = 1,performance = TRUE,maximize = NULL) {
+
+  if (is.na(metric)) {
+    stop("Please specify a character value for metric to get the best score/params...")
+  }
+
+  metric = tolower(metric)
```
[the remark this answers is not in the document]
fahadshery: You mentioned in your previous 2nd comment to remove the line? However, I am keeping it as your suggestion here.

R/select_best.R
fahadshery: or we could simply have an additional option of `sort = asc/desc` when returning the results? this way we wouldn't need it?

topepo: Doesn't `maximize` take the place of `sort`?

topepo: Also, I think that we'd rather avoid using almost right arguments like `sort == "asc" | sort == "ascending"` or `tolower(metric)`. I used to do that in `caret` and found it to be more trouble than it is worth. It also makes people think that all of the other arguments have shortcuts too.

fahadshery:
I have updated `tests` etc. but still not sure how to check `maximize` to do `ascending` or `descending` sort in the mean time before we implement it properly?? Could you please have a look at the new code and make suggestions if there are any other issues? Sorry for being a pain. Thank you for your patience.

topepo:
>  Sorry for being a pain. Thank you for your patience.

No worries. This is how it usually works

> I have updated tests etc. but still not sure how to check `maximize` to do ascending or descending sort in the mean time before we implement it properly?

I realize that I was unclear. `maximize` should stay a logical and default to `TRUE`. I would just test that you get the appropriate results for either of those two possible values. 

(We will eventually add an attribute to the grid object that has a table that has the appropriate value of `maximize` for each performance metric that was requested. At that point, we can remove the `maximize` argument. )

fahadshery:
ok. thank you for the explanation. I have made the changes. (Sorry had to pull before push again. )

topepo:
Merging but I might tweak the code a little bit later. 

Thanks a lot!

fahadshery:
It was an honour working with my `idol` in modelling (@topepo ) :D Thank you for the opportunity. What happens now? I am getting an option to `Delete fork` is this what happens after the merge normally?

topepo:
Thanks for your help!

You can delete the fork if you want. If you want to contribute more at a later date, you could keep it and update from the master branch (before starting another PR) to stay current.

topepo:
Just an FYI... I was adding `select_best()` to our training notes and I think that I will modify it to have `select_best()` choose the single best model (with no performance values) and `show_best()` which will show `top_n` and keep the performance columns.

PR #170, opened 2020-02-13
New argument order for tune_grid() and friends, to align better with fit()

juliasilge:
Check out this first idea 💡 for handling the arguments for `tune_grid()`, so that both this function and parsnip's `fit()` take, in this order:

- a model specification
- a formula, or recipe here (I called this argument `preprocess` in this PR but of course open to discuss)
- the "data", in this case the resamples

I updated the tests so that this will pass R CMD check but did not change anything in vignettes or README, until we decide if we like this.


topepo:
I was going to suggest that the message should show an example of what it should be. This _might_ be able to do it: 

```r
deprecate_msg <- function(cl) {
  func_sig <- rlang::expr_text(cl[1])
  # re-write call to switch arguments
  cl[[1]] <- rlang::expr(tune)
  pp <- cl[[2]]
  cl[[2]] <- cl[["model"]]
  cl[["model"]] <- pp
  names(cl)[1:3] <- ""
  msg <-
    paste(
      "The first argument to `tune_grid()` should be either a model or a",
      "workflow. In the future, you can use:\n",
      rlang::expr_text(cl)
    )
  lifecycle::deprecate_soft("0.0.2", what = func_sig, details = msg)
}

tune_grid.recipe <- function(object, model, resamples, ..., param_info = NULL,
                             grid = 10, metrics = NULL, control = control_grid()) {
  
  deprecate_msg(match.call())
  
  empty_ellipses(...)
  
  tune_grid(model, preprocessor = object, resamples = resamples,
            param_info = param_info, grid = grid,
            metrics = metrics, control = control)
}
```

However, called within `tune_grid.recipes()` or `tune_grid.formula()` the message doesn't print. 

``` r
library(tidymodels)
#> ── Attaching packages ──────────────────────────────────────────────────── tidymodels 0.0.4 ──
#> ✓ broom     0.5.4          ✓ recipes   0.1.9     
#> ✓ dials     0.0.4.9000     ✓ rsample   0.0.5     
#> ✓ dplyr     0.8.4          ✓ tibble    2.1.3     
#> ✓ ggplot2   3.2.1          ✓ tune      0.0.1.9000
#> ✓ infer     0.5.1          ✓ workflows 0.1.0     
#> ✓ parsnip   0.0.5          ✓ yardstick 0.0.5     
#> ✓ purrr     0.3.3
#> ── Conflicts ─────────────────────────────────────────────────────── tidymodels_conflicts() ──
#> x purrr::discard()    masks scales::discard()
#> x dplyr::filter()     masks stats::filter()
#> x dplyr::lag()        masks stats::lag()
#> x ggplot2::margin()   masks dials::margin()
#> x recipes::step()     masks stats::step()
#> x recipes::yj_trans() masks scales::yj_trans()
```

``` r

set.seed(6735)
```

``` r

# tuning recipe parameters:

spline_rec <-
  step_ns(disp, deg_free = tune("disp")) %>%
  step_ns(wt, deg_free = tune("wt"))

lin_mod <-
  linear_reg() %>%
  set_engine("lm")

# manually create a grid
spline_grid <- expand.grid(disp = 2:5, wt = 2:5)

# Warnings will occur from making spline terms on the holdout data that are
# extrapolations.
spline_res <-
  tune_grid(spline_rec, lin_mod, resamples = folds, grid = spline_grid)
```

<sup>Created on 2020-02-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Maybe @lionel- has a suggestion. We could just have `deprecate_msg()` generate the message and put the call to `lifecycle::deprecate_soft()` outside of it. It would be good to know if I'm doing something wrong though.

topepo:
For the documentation, maybe leave the second argument name out (to make the change less jarring):

```r
# Warnings will occur from making spline terms on the holdout data that are
# extrapolations.
spline_res <-
  tune_grid(lin_mod, spline_rec, resamples = folds, grid = spline_grid)
spline_res
```

instead of

```r
# Warnings will occur from making spline terms on the holdout data that are
# extrapolations.
spline_res <-
  tune_grid(lin_mod, preprocessor = spline_rec, resamples = folds, grid = spline_grid)
spline_res
```
My feelings are not all that strong about this so it is just a suggestion.

juliasilge:
I implemented the little message with the suggested call in a way that we can reuse the function for all the other tuning functions we are about to change as well (`fit_resamples()`, etc). Check it and see if you think that's a good idea.

Given that, do you want it in its own file? How do you all like to handle things like that style-wise?

``` r
library(tidymodels)
#> ── Attaching packages ────────────────────────────────────────────────────────────────────────────────────────────────────────── tidymodels 0.0.4 ──
#> ✓ broom     0.5.4          ✓ recipes   0.1.9     
#> ✓ dials     0.0.4          ✓ rsample   0.0.5     
#> ✓ dplyr     0.8.4          ✓ tibble    2.1.3     
#> ✓ ggplot2   3.2.1          ✓ tune      0.0.1.9000
#> ✓ infer     0.5.1          ✓ workflows 0.1.0.9000
#> ✓ parsnip   0.0.5          ✓ yardstick 0.0.5.9000
#> ✓ purrr     0.3.3
#> ── Conflicts ───────────────────────────────────────────────────────────────────────────────────────────────────────────── tidymodels_conflicts() ──
#> x purrr::discard()    masks scales::discard()
#> x dplyr::filter()     masks stats::filter()
#> x dplyr::lag()        masks stats::lag()
#> x ggplot2::margin()   masks dials::margin()
#> x recipes::step()     masks stats::step()
#> x recipes::yj_trans() masks scales::yj_trans()

folds <- vfold_cv(mtcars, v = 3)

svm_mod <-
  svm_rbf(cost = tune(), rbf_sigma = tune()) %>%
  set_engine("kernlab") %>%
  set_mode("regression")

tune_grid(mpg ~ ., svm_mod, resamples = folds, grid = 7)
#> Warning: `tune_grid.formula()` is deprecated as of lifecycle 0.0.2.
#> The first argument to `tune_grid()` should be either a model or a workflow. In the future, you can use:
#> tune_grid(svm_mod, mpg ~ ., resamples = folds, grid = 7)
#> This warning is displayed once per session.
#> Call `lifecycle::last_warnings()` to see where this warning was generated.
#> #  3-fold cross-validation 
#> # A tibble: 3 x 4
#>   splits          id    .metrics          .notes          
#> * <list>          <chr> <list>            <list>          
#> 1 <split [21/11]> Fold1 <tibble [14 × 5]> <tibble [0 × 1]>
#> 2 <split [21/11]> Fold2 <tibble [14 × 5]> <tibble [0 × 1]>
#> 3 <split [22/10]> Fold3 <tibble [14 × 5]> <tibble [0 × 1]>
```

<sup>Created on 2020-02-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

topepo:
This api change is the right thing to do but, I have to be honest

```r
tune_grid(svm_mod, mpg ~ ., resamples = folds, grid = 7)
```

makes me 🤢 It's better when piped so lets never do ☝️ in materials

lionel-:
@topepo What happens when you set `options(lifecycle_verbosity = "warning")`?

Soft-deprecation messages are only shown when called from the global environment or within package tests. `deprecate_soft()` takes an `env` argument which you can pass when you're calling it deep in your own stack. By default it looks in the grand-parent caller.

topepo:
@juliasilge I would put it in its own file (and thanks for seeing my awful bug in the original code). I'd go ahead and make the changes to the other functions based on this. 

@lionel- Using that option didn't change anything. Julia's edit works fine so we'll go with that. Thanks!

juliasilge:
I have now implemented this change (align the user-facing interface for arguments more closely with parsnip's `fit()`) for:

- `tune_grid()`
- `tune_bayes()`
- `fit_resamples()`
- `last_fit()`

I have updated for this change in the tests and vignettes. Take a look 👀 and see if I missed anything or if there are further changes that need to be made!

R/resample.R
```diff
@@ -95,19 +94,48 @@ fit_resamples.formula <- function(formula,
...
+#' @export
+#' @rdname fit_resamples
+fit_resamples.model_spec <- function(object,
+                                     preprocessor,
+                                     resamples,
+                                     ...,
+                                     metrics = NULL,
+                                     control = control_resamples()) {
+
+  if (is_missing(preprocessor) || !(inherits(preprocessor, "recipe") || inherits(preprocessor, "formula"))) {
```
topepo: Try to keep the lines within 80 char. You can use `styler` or cmd-shift-a. I often override those changes (just ask Davis) but keep the lines within the bounds.

R/resample.R
```diff
@@ -95,19 +94,48 @@ fit_resamples.formula <- function(formula,
                                   metrics = NULL,
                                   control = control_resamples()) {
 
+  lifecycle::deprecate_soft("0.0.2",
```
topepo: Can you make the deprecation deadline 0.1.2 since I'll be upping the version number to 0.1.0?

R/resample.R
```diff
@@ -5,7 +5,10 @@
 #' [tune_bayes()] for that), and is instead used for fitting a single
 #' model+recipe or model+formula combination across many resamples.
 #'
-#' @param object A workflow, formula, or recipe.
+#' @param object A `parsnip` model specification or a [workflows::workflow()].
```
topepo: Try using `#' @inheritParams` and fill it out once in `last_fit()`.

R/resample.R
lionel-: This is not a deadline but the date at which the function is deprecated. This should be the same version for the whole lifecycle of deprecation (soft-deprecated, deprecated, defunct).

juliasilge: For clarification here @topepo the next version on CRAN is going to be 0.1.0? Are we going from 0.0.1 straight to 0.1.0? I believe then this should be 0.1.0, because that is when the function is soft deprecated.

R/resample.R
juliasilge: Addressed in 9ca530512cb153a40454f439161d75f49180e167

R/resample.R
juliasilge: Addressed in 9ca530512cb153a40454f439161d75f49180e167

R/resample.R
juliasilge: I set this as 0.1.0 in 36bb83455820914d101b33f65c25bd8a12878e00. Let me know if I misinterpreted.

NEWS.md
```diff
@@ -1,9 +1,15 @@
 # tune 0.0.1.9000
 
+## Breaking Changes
+
+* The arguments to the main tuning/fitting functions (`tune_grid()`, 
+  `tune_bayes()`, etc) have been reordered to better align with parsnip's `fit()`.
```
DavisVaughan: Maybe another sentence or two: "Specifically, rather than having the recipe or formula as the first argument, you now pass the model first. The recipe and formula methods have been soft deprecated and will be removed in version ___"

R/last_fit.R
```diff
@@ -65,42 +69,66 @@ last_fit <- function(object, ...) {
...
   )
   rlang::abort(msg)
 }
 
 #' @rdname last_fit
 #' @export
 last_fit.recipe <- function(object, model, split, ..., metrics = NULL) {
+  lifecycle::deprecate_soft("0.1.0",
+                            what = "last_fit.recipe()",
+                            details = deprecate_msg(match.call(), "last_fit"))
```
DavisVaughan: What is the call manipulation getting us?

Would it be any simpler to do this?

```r
deprecate_msg <- function(what) {
  glue::glue(
    "As of tune 0.1.0, the first argument to `{what}()` ",
    "should be either a parsnip model or a workflow."
  )
}
```

Then use 

```r
deprecate_soft("0.1.0", what = "last_fit.recipe()", with = "last_fit.model_spec()", details = deprecate_msg("last_fit"))
```

R/last_fit.R
```diff
@@ -65,42 +69,66 @@ last_fit <- function(object, ...) {
...
+
+#' @export
+#' @rdname last_fit
+last_fit.model_spec <- function(object, preprocessor, split, ..., metrics = NULL) {
+
+  if (is_missing(preprocessor) ||
+      !(inherits(preprocessor, "recipe") ||
+        inherits(preprocessor, "formula"))) {
+    rlang::abort(paste("To tune a model spec, you must preprocess",
+                       "with a formula or recipe"))
```
DavisVaughan: Since we use them a lot, it might be useful to create a few helpers for these checks:

```r
is_recipe <- function(x) {
  inherits(x, "recipe")
}

is_preprocessor(x) {
  is_recipe(x) || is_formula(x) # is_formula() from rlang
}
```

Then this reduces to:

```r
if (is_missing(preprocessor) || !is_preprocessor(preprocessor)) {
  abort()
}
```

R/last_fit.R
```diff
@@ -65,42 +69,66 @@ last_fit <- function(object, ...) {
...
   }
 
-  workflow <- workflow()
-  workflow <- add_formula(workflow, formula)
-  workflow <- add_model(workflow, model)
+  empty_ellipses(...)
+
+  wflow <-
+    workflow() %>%
+    add_model(object)
```
DavisVaughan: Stylistically, I'm _personally_ team anti-pipe when I write package code unless it really helps. It has a very minor performance hit, can sometimes make the tracebacks a little harder to read if something does error, and for one-liners I often find the non-piped equivalent a little easier to read. I think Max feels differently though so its fine to keep as is

R/last_fit.R
```diff
@@ -65,42 +69,66 @@ last_fit <- function(object, ...) {
...
 
-  last_fit_workflow(workflow, split, metrics)
+  if (inherits(preprocessor, "recipe")) {
```
DavisVaughan: could use the `is_recipe()` and `is_formula()` helpers here

R/last_fit.R
```diff
@@ -65,42 +69,66 @@ last_fit <- function(object, ...) {
...
   msg <- paste0(
-    "The first argument to [last_fit()] should be either a ",
-    "formula, recipe, or workflow."
+    "The first argument to [last_fit()] should be either ",
+    "a model or workflow."
   )
   rlang::abort(msg)
 }
 
 #' @rdname last_fit
```
DavisVaughan: For the soft deprecated ones, maybe remove them from the rd file of `last_fit()`? That way new users never know they existed in the first place, because the methods themselves won't show up. 

That might mean you don't have to document `@param model` too

NEWS.md
DavisVaughan: Oooh I think you added the version note right after i added this review comment

R/last_fit.R
juliasilge: The call manipulation lets us give the user the new call they should use, [shown up here](https://github.com/tidymodels/tune/pull/170#issuecomment-587735689). It's not a huge deal either way and it would be simpler to not mess with the call, but my opinion is that it's nice to show the user what to do (can copy/paste the result we give them).

NEWS.md
juliasilge: Added more detail in 8b36a2f707eae569b252406191a4adebf76aa5eb

R/last_fit.R
DavisVaughan: Oh now I get it. Yea that seems useful!

R/last_fit.R
juliasilge: Addressed in db74821d2a0c096ca8f1075349acf74be8bb7dde

R/last_fit.R
juliasilge: I'm with you on this, especially for one-liners. Addressed in db74821d2a0c096ca8f1075349acf74be8bb7dde

R/last_fit.R
juliasilge: Addressed in db74821d2a0c096ca8f1075349acf74be8bb7dde

R/last_fit.R
juliasilge: Addressed in 987683ecd785220ca3048106b658dcc198c58dae

R/last_fit.R
juliasilge: I put these in a new `utils.R` file. Let me know if anybody has thought/opinions on that.

DavisVaughan: Makes sense to me!

topepo:
Oops. I hit the wrong review button; we're good

Issue #160, opened 2020-01-28
Parralel processing fails with function \"all_nominal\" (related to to issue #159)


kelseygonzalez:
Issued discovered during the Applied machine learning workshop. Related to issue #159 . 
suggested to submit issue from @jyuu . 

``` r
library(tidymodels)
#> Registered S3 method overwritten by 'xts':
#>   method     from
#>   as.zoo.xts zoo
#> -- Attaching packages ----------------------------------------------------------- tidymodels 0.0.3 --
#> v broom     0.5.3     v purrr     0.3.3
#> v dials     0.0.4     v recipes   0.1.9
#> v dplyr     0.8.3     v rsample   0.0.5
#> v ggplot2   3.2.1     v tibble    2.1.3
#> v infer     0.5.1     v yardstick 0.0.4
#> v parsnip   0.0.5
#> -- Conflicts -------------------------------------------------------------- tidymodels_conflicts() --
#> x purrr::discard()    masks scales::discard()
#> x dplyr::filter()     masks stats::filter()
#> x dplyr::lag()        masks stats::lag()
#> x ggplot2::margin()   masks dials::margin()
#> x recipes::step()     masks stats::step()
#> x recipes::yj_trans() masks scales::yj_trans()
library(tune)
library(doParallel)
#> Loading required package: foreach
#> 
#> Attaching package: 'foreach'
#> The following objects are masked from 'package:purrr':
#> 
#>     accumulate, when
#> Loading required package: iterators
#> Loading required package: parallel

data(Chicago)

us_hol <- 
  timeDate::listHolidays() %>% 
  stringr::str_subset("(^US)|(Easter)")

chi_rec <-
  recipe(ridership ~ ., data = Chicago) %>%
  step_holiday(date, holidays = us_hol) %>%
  step_date(date) %>%
  step_rm(date) %>%
  step_dummy(all_nominal()) %>%
  step_zv(all_predictors())

chi_folds <- rolling_origin(Chicago, initial = 364 * 15, assess = 7 * 4, skip = 7 * 4, cumulative = FALSE)

glmn_grid <- expand.grid(penalty = 10^seq(-3, -1, 
                                          length.out = 20), 
                         mixture = (0:5)/5)

glmn_rec <- chi_rec %>%  step_normalize(all_predictors()) 
glmn_mod <-
  linear_reg(penalty = tune(), mixture = tune()) %>% set_engine("glmnet")

ctrl <- control_grid(save_pred = TRUE,
[...]

Issue #159, opened 2020-01-28
Using parallel processing results in error with tune_grid()


pstraforelli:
As part of the Applied ML course, I have run into the following error when using parallel processing on a Windows machine.

```
library(tidymodels)
#> -- Attaching packages ---------------------------------------- tidymodels 0.0.3 --
#> v broom     0.5.4     v purrr     0.3.3
#> v dials     0.0.4     v recipes   0.1.9
#> v dplyr     0.8.3     v rsample   0.0.5
#> v ggplot2   3.2.1     v tibble    2.1.3
#> v infer     0.5.1     v yardstick 0.0.5
#> v parsnip   0.0.5
#> -- Conflicts ------------------------------------------- tidymodels_conflicts() --
#> x purrr::discard()    masks scales::discard()
#> x dplyr::filter()     masks stats::filter()
#> x dplyr::lag()        masks stats::lag()
#> x ggplot2::margin()   masks dials::margin()
#> x recipes::step()     masks stats::step()
#> x recipes::yj_trans() masks scales::yj_trans()
library(tune)
library(tidyverse)
library(lubridate)
#> 
#> Attaching package: 'lubridate'
#> The following object is masked from 'package:base':
#> 
#>     date
data(Chicago)
Chicago
#> # A tibble: 5,698 x 50
#>    ridership Austin Quincy_Wells Belmont Archer_35th Oak_Park Western Clark_Lake
#>        <dbl>  <dbl>        <dbl>   <dbl>       <dbl>    <dbl>   <dbl>      <dbl>
#>  1     15.7   1.46         8.37     4.60       2.01     1.42    3.32       15.6 
#>  2     15.8   1.50         8.35     4.72       2.09     1.43    3.34       15.7 
#>  3     15.9   1.52         8.36     4.68       2.11     1.49    3.36       15.6 
#>  4     15.9   1.49         7.85     4.77       2.17     1.44    3.36       15.7 
#>  5     15.4   1.50         7.62     4.72       2.06     1.42    3.27       15.6 
#>  6      2.42  0.693        0.911    2.27       0.624    0.426   1.11        2.41
#>  7      1.47  0.408        0.414    1.63       0.378    0.225   0.567       1.37
#>  8     15.5   0.987        4.81     3.52       1.34     0.879   1.94        9.02
#>  9     15.9   1.55         8.23     4.71       2.22     1.46    3.46       16.0 
[...]

Issue #157, opened 2020-01-25
add example using the with foreach::foreach() in tune_bayes


brunocarlin:
Can you guys show how to do it in the examples of the function?

topepo:
Can you be more specific? That function is used internally. Do you mean how to use parallel processing?

brunocarlin:
Yes for example I use

[cl <- makeForkCluster(16)
registerDoParallel(cl)
foreach::getDoParWorkers()](url)

topepo:
Forking doesn't work outside of linux and macOS. For Windows, you could use

```r
library(doParallel)
cl <- makePSOCKcluster(parallel::detectCores(logical = FALSE))
registerDoParallel(cl)
```
It is very much worth looking at the [`doFuture` package](https://cran.r-project.org/web/packages/doFuture/vignettes/doFuture.html). It has some nice features that work on top of `foreach`.

brunocarlin:
Really cool package, I will try it on windows tomorrow, if it works I will close the thread, if it doesn't I will try to make some reprex, sorry for the inconvenience it is so cool using this package

brunocarlin:
It seems Windowns fails somehow to use multiple workers

``` r
library(tidymodels)
#> Registered S3 method overwritten by 'xts':
#>   method     from
#>   as.zoo.xts zoo
#> -- Attaching packages ---------------------------------------------------- tidymodels 0.0.3 --
#> v broom     0.5.3     v purrr     0.3.3
#> v dials     0.0.4     v recipes   0.1.9
#> v dplyr     0.8.3     v rsample   0.0.5
#> v ggplot2   3.2.1     v tibble    2.1.3
#> v infer     0.5.1     v yardstick 0.0.4
#> v parsnip   0.0.5
#> -- Conflicts ------------------------------------------------------- tidymodels_conflicts() --
#> x purrr::discard()    masks scales::discard()
#> x dplyr::filter()     masks stats::filter()
#> x dplyr::lag()        masks stats::lag()
#> x ggplot2::margin()   masks dials::margin()
#> x recipes::step()     masks stats::step()
#> x recipes::yj_trans() masks scales::yj_trans()
library(tune)
library(kernlab)
#> 
#> Attaching package: 'kernlab'
#> The following object is masked from 'package:purrr':
#> 
#>     cross
#> The following object is masked from 'package:ggplot2':
#> 
#>     alpha
#> The following object is masked from 'package:scales':
#> 
#>     alpha

# Load data
[...]

PR #171, opened 2020-02-16
changes for #157 and #159 and #160

topepo:
Added a few model models to be fully loaded by `foreach`


topepo:
closes #157 

closes #159 

closes #160

Issue #172, opened 2020-02-17
eliminate redundant `maximize` argument


topepo:
`select_*()` functions have an argument called `maximize` which we don't need anymore. The current version of `yardstick` has a direction build into the metric set:

``` r
library(tidymodels)
#> ── Attaching packages ─────────────────────────────── tidymodels 0.0.4 ──
#> ✓ broom     0.5.4          ✓ recipes   0.1.9     
#> ✓ dials     0.0.4.9000     ✓ rsample   0.0.5     
#> ✓ dplyr     0.8.4          ✓ tibble    2.1.3     
#> ✓ ggplot2   3.2.1          ✓ tune      0.0.1.9000
#> ✓ infer     0.5.1          ✓ workflows 0.1.0     
#> ✓ parsnip   0.0.5          ✓ yardstick 0.0.5     
#> ✓ purrr     0.3.3
#> ── Conflicts ────────────────────────────────── tidymodels_conflicts() ──
#> x purrr::discard()    masks scales::discard()
#> x dplyr::filter()     masks stats::filter()
#> x dplyr::lag()        masks stats::lag()
#> x ggplot2::margin()   masks dials::margin()
#> x recipes::step()     masks stats::step()
#> x recipes::yj_trans() masks scales::yj_trans()

names(formals(select_best))
#> [1] "x"        "metric"   "maximize"
names(formals(select_by_one_std_err))
#> [1] "x"        "..."      "metric"   "maximize"
names(formals(select_by_pct_loss))
#> [1] "x"        "..."      "metric"   "maximize" "limit"

tune:::metrics_info(metric_set(rmse, rsq))
#> # A tibble: 2 x 3
#>   .metric direction type   
#>   <chr>   <chr>     <chr>  
#> 1 rmse    minimize  numeric
#> 2 rsq     maximize  numeric
```

<sup>Created on 2020-02-16 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

PR #174, opened 2020-02-17
out-of-sample prediction averaging

topepo:
The main changes here are: 

 * `tune_grid()`, `tune_bayes()`, `fit_resamples()`, and `last_fit()` save the parameter sets and metric sets as attributes. 

* `collect_predictions()` gains a `parameters` argument to pre-filter the results if you only what results back for one or two models. 

* `collect_predictions()` gains a `summarize` argument  that will average the out-of-sample predictions over resampling replicates. We do this by using `pivot_longer()` and averaging, the `pivot_wider()` to go back to the original data structure. I'll ask some reviewers if there is a better way of doing this.


romainfrancois:
I'd have to look in more details, but this feels like a use case for `across()` : 

``` r
library(dplyr, warn.conflicts = FALSE)

numeric_summarize <- function(x, wider = TRUE) {
  pred_cols <- grep("^\\.pred", names(x), value = TRUE)
  nms <- names(x)
  group_cols <- nms[!(nms %in% pred_cols)]
  x <-
    x %>%
    tidyr::pivot_longer(
      cols = c(dplyr::one_of(pred_cols)),
      names_to = ".column",
      values_to = ".value"
    ) %>%
    dplyr::group_by(!!!rlang::syms(group_cols), .column) %>%
    dplyr::summarize(.value = mean(.value, na.rm = TRUE))

  if (wider) {
    x <-
      x %>%
      dplyr::ungroup() %>%
      tidyr::pivot_wider(
        id_cols = c(dplyr::one_of(group_cols)),
        names_from = ".column",
        values_from = ".value"
      )
  }
  x
}

df <- tibble(x = sample(1:10, 1e6, replace = TRUE), .pred.a = rnorm(1e6), .pred.b = rnorm(1e6), .pred.c = rnorm(1e6))
bench::mark(
  one   = numeric_summarize(df, TRUE), 
  two   = df %>% group_by(x) %>% summarise(across(starts_with(".pred"), ~mean(., na.rm = TRUE)))
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 2 x 6
#> 1 one           710ms    710ms      1.41     357MB     8.45
#> 2 two           184ms    188ms      4.89     112MB     8.14
```

<sup>Created on 2020-02-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

R/collect.R
```diff
@@ -5,22 +5,42 @@
 #'  = TRUE` should have been used.
 #' @param summarize A logical; should metrics be summarized over resamples
 #' (`TRUE`) or return the values for each individual resample. Note that, if `x`
-#' is created by [last_fit()], `summarize` has not effect.
+#' is created by [last_fit()], `summarize` has not effect. For the other object
```
DavisVaughan: `summarize` has not effect -> `summarize` has no effect

R/collect.R
```diff
@@ -60,9 +84,217 @@ collect_predictions <- function(x) {
...
+    x <- collector(x, coll_col = coll_col)
+  }
+  x
+}
+
+filter_predictions <- function(x, parameters) {
+  if (is.null(parameters)) {
+    return(x)
+  }
+  metrics <- attr(x, "metrics")
```
DavisVaughan: Is `metrics` used other than to check if it exists? Does the check on `params` cover that?

R/collect.R
```diff
@@ -60,9 +84,217 @@ collect_predictions <- function(x) {
...
+  params <- attr(x, "parameters")
```
DavisVaughan: I'm a little bit worried about attributes being somehow dropped along the way after fitting but before `collect_predictions()` is called. I suppose it would be pretty rare

R/collect.R
```diff
@@ -60,9 +84,217 @@ collect_predictions <- function(x) {
...
+#  mode is used to summarize.
+
+numeric_summarize <- function(x, p, wider = TRUE) {
+  pred_cols <- grep("^\\.pred", names(x), value = TRUE)
+  nms <- names(x)
+  group_cols <- nms[!(nms %in% pred_cols)]
+  x <-
+    x %>%
+    tidyr::pivot_longer(
+      cols = c(dplyr::one_of(pred_cols)),
```
DavisVaughan: You might switch to tidyselect 1.0.0 and use `tidyselect::all_of()`, which is slightly more clear

R/collect.R
```diff
@@ -60,9 +84,217 @@ collect_predictions <- function(x) {
...
+        ),
+        collapse = "\n"
+      )
+    )
+
+    return(x)
+  }
+
+  param_names <- params$id
+  parameters <- dplyr::select(parameters, dplyr::one_of(param_names))
```
DavisVaughan: This could probably be `tidyselect::any_of()`.

R/collect.R
```diff
@@ -60,9 +84,217 @@ collect_predictions <- function(x) {
...
+  group_cols <- nms[!(nms %in% pred_cols)]
+  x <-
+    x %>%
+    tidyr::pivot_longer(
+      cols = c(dplyr::one_of(pred_cols)),
+      names_to = ".column",
+      values_to = ".value"
+    ) %>%
+    dplyr::group_by(!!!rlang::syms(group_cols), .column) %>%
+    dplyr::summarize(.value = mean(.value, na.rm = TRUE))
```
DavisVaughan: Should you `ungroup()` here? Right now you just `ungroup()` if `wider = TRUE`. If `wider = FALSE` this will likely return a grouped_df. Not sure if you want that

R/collect.R
```diff
@@ -60,9 +84,217 @@ collect_predictions <- function(x) {
...
+
+  avg_group_cols <- group_cols[group_cols != ".column"]
```
DavisVaughan: When would it ever contain `.column`?

R/collect.R
```diff
@@ -60,9 +84,217 @@ collect_predictions <- function(x) {
...
+  x <-
+    x %>%
+    dplyr::full_join(
+      x %>% dplyr::summarize(.totals = sum(.value)),
+      by = avg_group_cols
+    ) %>%
```
DavisVaughan: Is this just `mutate(.totals = sum(.value))`? Maybe I'm reading it wrong

R/collect.R
```diff
@@ -60,9 +84,217 @@ collect_predictions <- function(x) {
...
+    dplyr::mutate(.value = .value/.totals) %>%
+    dplyr::select(-.totals)
+
+  if (wider) {
+    x <-
+      x %>%
+      dplyr::ungroup() %>%
```
DavisVaughan: Same comment about placement of `ungroup()`

R/collect.R
DavisVaughan: Hmm maybe you do want it for `prob_and_class_summarize()` later on?

R/collect.R
```diff
@@ -60,9 +84,217 @@ collect_predictions <- function(x) {
...
+  x
+}
+
+class_summarize <- function(x, p) {
+  pred_cols <- grep("^\\.pred", names(x), value = TRUE)
+  nms <- names(x)
+  group_cols <- nms[!(nms %in% pred_cols)]
+  outcome_col <- group_cols[!(group_cols %in% c(p, ".row", ".iter"))]
+  lvl <- levels(x[[outcome_col]])
+  ord <- is.ordered(x[[outcome_col]])
```
DavisVaughan: Do you need these?

R/collect.R
```diff
@@ -60,9 +84,217 @@ collect_predictions <- function(x) {
...
+  x
+}
+
+
+average_predictions <- function(x, grid = NULL) {
+  metric_types <- metrics_info(attr(x, "metrics"))$type
+  param_names <- attr(x, "parameters")$id
+
+  if (!is.null(grid)) {
+    grid <- dplyr::select(grid, dplyr::one_of(param_names))
```
DavisVaughan: `tidyselect::any_of()` if you want to

R/collect.R
```diff
@@ -60,9 +84,217 @@ collect_predictions <- function(x) {
...
+
+  x <-
+    x %>%
+    collect_predictions() %>%
+    dplyr::select(-starts_with("id"))
+
+  if (all(metric_types == "numeric")) {
+    x <- numeric_summarize(x, param_names)
+  } else {
+    if (all(metric_types == "prob")) {
```
DavisVaughan: `else if (all(metric_types == "prob")) {`? rather than nesting

R/collect.R
```diff
@@ -60,9 +84,217 @@ collect_predictions <- function(x) {
...
+      x <- prob_summarize(x, param_names)
+    } else {
+      if (all(metric_types == "class")) {
```
DavisVaughan: Same here about `else if`

R/collect.R
topepo: I'll run some tests to make sure that the `dplyr` verbs don't do that. Can you think of another way to enbed them?

R/collect.R
topepo: That's not in the CRAN version right now (right?)

R/collect.R
topepo: I'll want to get another version of `tune` out soon. What is the expected release time for `tidyselect`?

R/collect.R
topepo: Yes. Otherwise I group again there.

R/collect.R
topepo: Good catch. That's form some vestigial code.

R/collect.R
topepo: Yes since I'll be creating a new factor later and I want it to match the observed classes.

R/collect.R
topepo: All good _except_ `mutate.rset` removes them 😞

topepo: fixed

R/collect.R
topepo: nvmd; I was looking at the wrong function; yes, those are also vestigial.

R/collect.R
topepo: It looks like it is; The Google lied!

image

juliasilge:
I will admit that the motivation for saving the parameters and metric class/direction as attributes is not transparent to me. 🙈 Does this set up some planned functionality? Just important to have a record of exactly what kind of tuning was done?

topepo:
For functionality. 

There were/are a few things that we want to do with the results (like in the `select_*` functions where we need to subset on parameters. The current code does some guesswork to determine what column were parameters etc. This makes it easier and less likely to break. 

For the metrics, there is #172 which would require having the metric set around after the tune/resampling are finished.

R/collect.R
```diff
@@ -5,22 +5,42 @@
 #'  = TRUE` should have been used.
 #' @param summarize A logical; should metrics be summarized over resamples
 #' (`TRUE`) or return the values for each individual resample. Note that, if `x`
-#' is created by [last_fit()], `summarize` has not effect.
+#' is created by [last_fit()], `summarize` has no effect. For the other object
+#' types, the method of summarizing predictions is detailed below.
+#' @param parameters An optional tibble of tuning parameter values that can be
+#'  used to filter the predicted values before processing. This tibble should
+#'  only have columns for each tuning parameter identifier (e.g. `"my_param"`
+#'  if `tune("my_param")` was used).
```
juliasilge: Can there be a more transparent example of how to use this? Either literally a whole entry in `@examples` (maybe switch from `fit_resamples()` to tune_grid()`?) or else change the documentation here to something like:
```
This tibble should only have columns for tuned parameters (e.g., if `my_param = tune()` was used with tuning values of 3, 5 and 10, a possible value is `parameters = tibble(my_param = 3).` 
```

topepo:
@juliasilge This comment of your has disappeared: 

> Can there be a more transparent example of how to use this? Either literally a whole entry in `@examples` (maybe switch from `fit_resamples()` to `tune_grid()`?) or else change the documentation here to something like:

> This tibble should only have columns for tuned parameters (e.g., if `my_param = tune()` was used with tuning values of 3, 5 and 10, a possible value is `parameters = tibble(my_param = 3).` 

I updated the examples in 30b9c9ab9 to make this better.

R/collect.R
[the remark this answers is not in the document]
juliasilge: Addressed in 30b9c9ab99cf4680fb1edb64af7fe01441152465

R/collect.R
```diff
@@ -60,9 +84,210 @@ collect_predictions <- function(x) {
...
+      x %>%
+      dplyr::ungroup() %>%
+      tidyr::pivot_wider(
+        id_cols = c(dplyr::one_of(group_cols)),
+        names_from = ".column",
+        values_from = ".value"
+      )
+  }
+  x
+}
```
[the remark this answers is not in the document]
juliasilge: Addressed in 30b9c9ab99cf4680fb1edb64af7fe01441152465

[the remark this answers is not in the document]
romainfrancois: oh 😬 sorry I live in master dplyr, another option is `summarise_at()` : 

``` r
library(dplyr, warn.conflicts = FALSE)
df <- tibble(x = sample(1:10, 1e6, replace = TRUE), .pred.a = rnorm(1e6), .pred.b = rnorm(1e6), .pred.c = rnorm(1e6))
df %>% 
  group_by(x) %>% 
  summarise_at(vars(starts_with(".pred")), ~mean(., na.rm = TRUE))
#> # A tibble: 10 x 4
#>        x   .pred.a   .pred.b   .pred.c
#>    <int>     <dbl>     <dbl>     <dbl>
#>  1     1  0.00357  -0.000707  0.000568
#>  2     2 -0.00669  -0.00135   0.000644
#>  3     3 -0.00215  -0.00220  -0.00130 
#>  4     4 -0.000861  0.00266  -0.00363 
#>  5     5  0.000988  0.00309  -0.00732 
#>  6     6  0.00608  -0.00191  -0.000505
#>  7     7 -0.00153   0.00223  -0.00500 
#>  8     8 -0.00440   0.00122  -0.00374 
#>  9     9  0.00128  -0.00105  -0.00926 
#> 10    10 -0.00530  -0.00180   0.00266
```

<sup>Created on 2020-02-20 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9000)</sup>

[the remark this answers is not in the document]
topepo: Changes based on this and the discussion at the group meeting today are in f644cabdc

PR #176, opened 2020-02-21
No more `maximize`. Closes #172.

### LINKED ISSUE #172: eliminate redundant `maximize` argument

juliasilge:
There's a new little function `is_metric_maximize()` that gets called before `check_metric_choice()` now.


R/select_best.R
```diff
@@ -38,7 +37,7 @@
 #' \donttest{
 #' data("example_ames_knn")
 #'
-#' show_best(ames_iter_search, metric = "rmse", maximize = FALSE)
+#' show_best(ames_iter_search, metric = "rmse")
```
topepo: Can you add `...` and, if someone passes `maximize`, write out a message that it's no longer needed? I don't think that we need to use the lifecycle code for this since it is a minor function.

topepo:
Please add a note to the NEWS file.

R/select_best.R
juliasilge: This will involve moving the dots to the end of the function call for `select_by_pct_loss()` and `select_by_one_std_err()`, I believe. That's OK? Looks OK to me.

juliasilge: Nope, never mind. Fine the way it is.

juliasilge:
OK, I think this is *actually* ready for review now @topepo. 💃

topepo:
Looks good with one suggested change. I ran the conf notes through this version and I think that the message(which I thought was fine before) is a little "inside baseball". Can you change: 

> Ignored the `maximize` argument. Direction is built in to each metric set.

to something like 

> The `maximize` argument is no longer needed. This value was ignored. 

After that, we are 💯

Issue #183, opened 2020-03-10
control_resamples doesn't work as other control_*


cimentadaj:
First, thanks for `tidymodels`! You're open source work is very valuable to the community. 

## The problem

I was trying to use `tune::control_resamples()` similarly as `parsnip::control_parsnip()` but noticed that none of the `tune::control_*` functions have a custom class similar to `parsnip::control_parsnip()` nor a custom printout.

Have you considered making them work similarly? 

## Reproducible example

```r
library(parsnip)
library(tune)

control_parsnip()
#> parsnip control object
class(control_parsnip())
#> [1] "control_parsnip"

control_resamples()
#> $verbose
#> [1] FALSE
#> 
#> $allow_par
#> [1] TRUE
#> 
#> $extract
#> NULL
#> 
#> $save_pred
#> [1] FALSE
#> 
#> $pkgs
#> NULL
class(control_resamples())
#> [1] "list"

control_grid()
#> $verbose
#> [1] FALSE
#> 
#> $allow_par
#> [1] TRUE
#> 
#> $extract
#> NULL
#> 
#> $save_pred
#> [1] FALSE
#> 
#> $pkgs
#> NULL
class(control_grid())
#> [1] "list"
```

If you're interested I could send a PR with the same structure as `parsnip::control_parsnip()`.

topepo:
A PR would be great.

topepo:
Thanks!

PR #195, opened 2020-04-03
Adds custom class for control_* funs

cimentadaj:
This PR implements what was discussed in #183 .

* Adds a custom class to `control_bayes()` and checks the class in the tests

* Adds a custom class to `control_grid()` of `control_grid` and `control_resamples`. Not sure whether that makes sense since every call to `val_class_*` explicitly calls `control_grid` even if it was called from `control_resamples`. Should `control_resamples` be a separate copy of `control_grid` w/ same args, etc..?

* Implements a simple printout _a la_ `control_parsnip` for the three `control_*`.

Let me know your thoughts on what is useful or what can be improved.

### DIFF
--- a/R/control.R
+++ b/R/control.R
@@ -40,8 +40,20 @@ control_grid <- function(verbose = FALSE, allow_par = TRUE,
   val_class_or_null(extract, "function", "control_grid()")
 
 
-  list(verbose = verbose, allow_par = allow_par, extract = extract,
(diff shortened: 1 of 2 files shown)


cimentadaj:
Tests are not passing but that's independent of these additions.

topepo:
CI tests are currently borked until the new `dials` (0.0.6) has binaries built 😞

PR #219, opened 2020-05-28
Changes for new rsample/dplyr/dials versions

topepo:
`rset` attributes are saved directly. 

Print method was changed to accommodate this change. 

Test objects were recreated using current devel versions. 

Potentially requires more changes for `dplyr` compatibility. @DavisVaughan


R/resample.R
```diff
@@ -139,6 +139,9 @@ resample_workflow <- function(workflow, resamples, metrics, control) {
 
   has_formula <- has_preprocessor_formula(workflow)
 
+  resample_info <- pull_rset_attributes(resamples)
+  resamples <- tibble::as_tibble(resamples)
```
DavisVaughan, line 143: we might should go ahead and use something like this to drop rsample attributes

```
new_bare_tibble <- function(x) {
  x <- new_data_frame(x)
  new_tibble(x, nrow = nrow(x))
}
```

R/tune_results.R
```diff
@@ -15,6 +15,24 @@ print.tune_results <- function(x, ...) {
...
+    cat("# Resampling results\n")
+  } else {
+    cat("# Tuning results\n")
+  }
+
+  att <- attributes(x)
+  if (any(names(att) == "rset_info")) {
+    cat("#", att$rset_info$label, "\n")
+  } else {
+    if (inherits(x, "rset")) {
```
DavisVaughan, line 29: It should never inherit from rset anymore right? Is this to be backwards compatible? If so, I would extract it into a helper called `print_tune_results_compat()` or something to make it super clear

.github/workflows/R-CMD-check.yaml
```diff
@@ -17,7 +17,8 @@ jobs:
         config:
           - {os: windows-latest, r: '3.6'}
           - {os: windows-latest, r: 'devel'}
-          - {os: macOS-latest, r: '3.6'}
+          - {os: macOS-latest, r: '3.6', dev_dplyr: true}
+          - {os: macOS-latest, r: '3.6', dev_dplyr: false}
```
DavisVaughan, line 21: At this point I think we can have 3 variants of mac 3.6 for:

```
- {os: macOS-latest, r: '3.6', old_dplyr: true,  dev_tidymodels: true}
- {os: macOS-latest, r: '3.6', old_dplyr: true,  dev_tidymodels: false}
- {os: macOS-latest, r: '3.6', old_dplyr: false, dev_tidymodels: true}
- {os: macOS-latest, r: '3.6', old_dplyr: false, dev_tidymodels: false}
```

which later we can reduce to

```
- {os: macOS-latest, r: '3.6', old_dplyr: true}
- {os: macOS-latest, r: '3.6', old_dplyr: false}
```

R/extract.R
```diff
@@ -40,7 +40,7 @@ pulley <- function(resamples, res, col) {
     pulled_vals <- tidyr::nest(pulled_vals, -starts_with("id"), .key = !!col)
   }
 
-  res <- full_join(resamples, pulled_vals, by = id_cols)
+  res <- full_join(tibble::as_tibble(resamples), pulled_vals, by = id_cols)
```
DavisVaughan, line 43: Could use `new_bare_tibble()` here too

R/tune_results.R
topepo: The print method works for older objects and I think that is a good idea. The helpers should work too. 

`print_tune_results_compat()` is fine with me.

PR #230, opened 2020-06-17
first pass at a filtering function + no irises

topepo:
This function can remove some tuning parameter results form `.metrics`, `.predictions`, and `.extracts`. 

@DavisVaughan and @juliasilge: any thoughts on the function name?


juliasilge:
I do feel a little uncertain about `filter_` as a name given that that there are rows _and_ columns in the tibble going in as a parameter. I'm not 100% on board, I don't think. Maybe let's get a little more input on options.

PR #258, opened 2020-08-07
Correct .config qualitative identifier for predictions in submodel case

juliasilge:
The `.config` qualitative identifier was being incorrectly matched up 🐛 to predictions because of how I set up `cbind()` in `append_predictions()`; the predictions in the submodel case do not come out of `predict_model()` in a very predictable fashion so we can't count on that. 

This PR instead creates a little dataframe of params and the `.config` identifier and uses an `inner_join()` instead.

Note that the dataframe is created using `across()` because we don't know the column names ahead of time:

```r
dplyr::distinct(metrics, dplyr::across(c(param_names, ".config")))
```

I don't know if we think it's too soon to use something from dplyr 1.0.0 in this package.

This PR addresses tidymodels/stacks#29. We know have unique but more importantly _correct_ qualitative identifiers.

``` r
library(tidymodels)
set.seed(1234)

mtcars_tb <- mtcars %>% 
  as_tibble() %>% 
  mutate(vs = c(sample(vs, 0), rep(NA_integer_, 10), sample(vs, 22)),
         disp = c(sample(disp, 22), rep(NA_integer_, 10)))

train_test_split <- initial_split(mtcars_tb)
train <- training(train_test_split)
test <- testing(train_test_split)

cv_fold_mtc <- vfold_cv(mtcars_tb)

recipe_naomit <- recipe(mpg ~ disp + vs, data = train) %>% 
  step_naomit(all_predictors(), skip = TRUE) %>%
  step_normalize(all_numeric()) 

lasso_mod <- 
  linear_reg() %>%
  set_engine("glmnet") %>% 
  set_args(penalty = tune(),
           mixture = 1)

wflow <-
  workflow() %>% 
  add_recipe(recipe_naomit) %>%
  add_model(lasso_mod)

lasso_tune <- 
  tune_grid(
    object = wflow, 
    resamples = cv_fold_mtc, 
    grid = 10,
    control = control_grid(save_pred = TRUE)
  )

collect_predictions(lasso_tune, summarize = TRUE) %>%
  count(.row, .config) %>%
  filter(n > 1)
#> # A tibble: 0 x 3
#> # … with 3 variables: .row <int>, .config <chr>, n <int>
```

<sup>Created on 2020-08-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>


topepo:
Would this lead to a `dplyr` version requirement for >= 1.0.0?

PR #260, opened 2020-08-09
Small typo in tune_grid() documentation

hnagaty:
I found 2 small typos in the documentation of `tune_grid()` and I fixed them here.

Those are:
1- For example, if a parameter is marked for optimization using penalty = tune(), there should be a column **names tune**. ==> there should be a column **named penalty**
2- For example, if the number of PCA components in a recipe step are being tuned over three values (along with model tuning parameters), only three recipes **are are** trained ==> double are


juliasilge:
Thank you so much @hnagaty! 🙌

PR #273, opened 2020-09-03
Construct a "manual" rset for usage in `last_fit()`

### LINKED ISSUE #264: split_to_rset() uses randomness when it doesn't need to
Extracted from https://github.com/tidymodels/tune/issues/11#issuecomment-675648443, because I now believe this is a separate standalone issue.

`split_to_rset()` calls `rsample::mc_cv()` just to get the structure of that rset subclass, but then we override the results with `x`. HOWEVER, it relies on randomness to generate the object that we overwrite. This causes some confusion with reproducibility between `last_fit()` and direct usage of `fit()/predict()`.

```r
> tune:::split_to_rset
function (x) 
{
    prop <- length(x$in_id)/nrow(x$data)
    res <- rsample::mc_cv(x$data, times = 1, prop = prop)
    res$splits[[1]] <- x
    res
}
```

Since we really just need to generate the structure the rset object, we should consider using the below rewrite instead. It is still suboptimal because it hardcodes the structure of an mc_cv rset, but I can live with that.

```r
# Manually construct an `mc_cv()` rset.
# Don't call `mc_cv()` directly, as that will mess with the random seed
split_to_rset <- function(x) {
  times <- 1L
  prop <- length(x$in_id) / nrow(x$data)
  strata <- FALSE
  
  attrib <- list(prop = prop, times = times, strata = strata)
  
  splits <- list(x)
  
  ids <- "Resample1"
  
  rsample::new_rset(
    splits = list(x), 
    ids = ids,
    attrib = attrib,
    subclass = c("mc_cv", "rset")
  )
}
```

DavisVaughan:
Closes #264 

Rather than using `mc_cv()`, this now constructs a new "manual" rset object from a set of splits/ids (here, just 1 split and 1 id). This avoids the randomness issue altogether.

I think it might be nice to move this manual rset idea to rsample eventually. It would be a simple wrapper around `new_rset()` that also has the vctrs/dplyr support built in, but wouldn't carry around any extra attributes. People could use `make_splits()` to construct their splits using manual indices, then pass the splits to `manual_rset()`. If you like that idea, I could move these pieces to rsample now and then make tune import dev rsample.


topepo:
> If you like that idea, I could move these pieces to rsample now and then make tune import dev rsample.

That's a good idea. Go ahead and do that.

PR #274, opened 2020-09-10
Add `workflows::add_variables()` support

DavisVaughan:
Requires dev workflows

This PR adds support for `add_variables()` from workflows. This is a "new" preprocessor type, so it requires its own code paths. This makes things a little uglier in the short term, but I'm fairly optimistic that this can be collapsed back down in a future PR that tries to unifying the code paths.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -30,7 +30,7 @@ Imports:
     GPfit,
     foreach,
     parsnip (>= 0.0.4),
-    workflows (>= 0.1.0),
+    workflows (>= 0.1.3.9000),
     hardhat (>= 0.1.0),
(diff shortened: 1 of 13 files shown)


topepo (changes requested):
It looks good (knowing that we will clean up later)! 

Can you update the news file and add an example to the `tune_grid()` and `fit_resamples()` docs (just using `lm()` and `mtcars` as-is)?

topepo:
Also, I just fixed a `dials` bug that will make CI fail (tidymodels/dials#143). You probably want to add `dials (>= 0.0.8.9001)` and a remotes for `dials`

PR #279, opened 2020-09-19
Fix comparison with NA (produced by metrics due lack of data)

dmalkr:
When improved metrics returns `NA`, `tune_bayes` fail with this message:

```
Error in if (is_better) { : missing value where TRUE/FALSE needed
Calls: %>% ... tune_bayes.workflow -> tune_bayes_workflow -> update_score_card
```

So it is need to check `is_better` value.


topepo:
Is there a small reproducible example to add as a test case?

PR #286, opened 2020-09-28
Option to save GP results to tempdir()

topepo:


topepo:
Small change; it now saves the model object instead of a grid of predictions.

_pkgdown.yml
```diff
@@ -40,6 +40,7 @@ reference:
     - parameters.workflow
     - tunable
     - tune_args
+    - message_wrap
```
DavisVaughan, line 43: What is this?

tests/testthat/test-bayes.R
```diff
@@ -385,6 +386,10 @@ test_that("retain extra attributes", {
   expect_true(inherits(att$parameters, "parameters"))
   expect_true(inherits(att$metrics, "metric_set"))
 
+  files <- list.files(path = tempdir(), pattern = "^gp_candidates")
```
DavisVaughan, line 389: Are you allowed to write to the tempdir when CRAN checks your package? Do you have to clean up after yourself? I'm not sure of the rules here but it might be worth checking

DavisVaughan (approved):

_pkgdown.yml
topepo: I wrote a function that uses `message()` but respects the current page width.

topepo: https://github.com/tidymodels/tune/blob/246983accbe756c9044138096d99067fa20413f9/R/logging.R#L29

I had not listed it on the `pkgdown` site.

tests/testthat/test-bayes.R
EmilHvitfeldt: I have at least one package that has a function that writes to `tempfile()` and they passed CRAN checks. Don't know if there is a fundamental difference between `tempfile()` and `tempdir()`.

https://github.com/EmilHvitfeldt/wordsalad/blob/master/R/fasttext.R#L67

topepo: `tune` doesn't run tests on CRAN but win-builder did not flag the use of `tempdir()` so I'll go with it.

Issue #220, opened 2020-05-29
tune_bayes() link doesn't work in pkgdown


EmilHvitfeldt:
https://tune.tidymodels.org/reference/tune_bayes.html
<img width="1210" alt="Screen Shot 2020-05-29 at 4 13 21 PM" src="https://user-images.githubusercontent.com/14034784/83312501-5a134e80-a1c7-11ea-8911-8ce7cc9958c5.png">

juliasilge:
This is because of the file being called `tune_Bayes.R`.

topepo:
This is changed in #219

juliasilge:
We had a CRAN release 6 days ago and this is still broken. The file is still called `tune_Bayes.html` instead of `tune_bayes.html`, which you [can see here](https://github.com/tidymodels/tune/tree/master/docs). I'm going to mark this as something to fix for the next release.

juliasilge:
Closed in #290

PR #290, opened 2020-10-01
Set up automatic pkgdown

juliasilge:
This PR sets up automatic pkgdown for tune. I suggest that we **not** use:

```yml
development:
  mode: auto
```

for now until the next release, because of the problem with the issue template; let's just get the pkgdown site up and working on the devel version for now.

I also fixed up the `CONTRIBUTING` with more info on how to fork and branch, like @DavisVaughan suggested on another repo.


topepo:
This will probably solve #220

PR #294, opened 2020-10-06
Set final workflow in last_fit() to `trained`

juliasilge:
The objects that end up in the `.workflow` column of `last_fit()` output currently have `trained` as false, meaning you can't predict on them. This PR sets:

```r
res$.workflow[[1]]$trained <- TRUE
```
in the same way that `fit.workflow` does, so that these workflows can be used for prediction.

``` r
library(tidymodels)

set.seed(6735)
tr_te_split <- initial_split(mtcars)

lin_mod <- linear_reg() %>%
  set_engine("lm")

car_wf <-
  workflow() %>%
  add_formula(mpg ~  .) %>%
  add_model(lin_mod)

car_fit <- last_fit(car_wf, split = tr_te_split)

predict(car_fit$.workflow[[1]], new_data = training(tr_te_split))
#> # A tibble: 24 x 1
#>    .pred
#>    <dbl>
#>  1  21.9
#>  2  22.4
#>  3  25.4
#>  4  21.6
#>  5  21.9
#>  6  13.3
#>  7  24.5
#>  8  17.6
#>  9  19.4
#> 10  13.7
#> # … with 14 more rows
```

<sup>Created on 2020-10-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

### DIFF
--- a/R/last_fit.R
+++ b/R/last_fit.R
@@ -136,6 +136,7 @@ last_fit_workflow <- function(object, split, metrics) {
       control = ctrl
     )
   res$.workflow <- res$.extracts[[1]][[1]]
+  res$.workflow[[1]]$trained <- TRUE
   res$.extracts <- NULL
   class(res) <- c("last_fit", class(res))
   class(res) <- unique(class(res))
--- a/R/last_fit.R
+++ b/R/last_fit.R
@@ -136,7 +136,6 @@ last_fit_workflow <- function(object, split, metrics) {
       control = ctrl
     )
   res$.workflow <- res$.extracts[[1]][[1]]
-  res$.workflow[[1]]$trained <- TRUE
   res$.extracts <- NULL
   class(res) <- c("last_fit", class(res))
   class(res) <- unique(class(res))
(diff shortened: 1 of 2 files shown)


DavisVaughan:
I think this is an issue with workflows. If you fit the workflow with `fit()`, then this method runs which sets `$trained <- TRUE`, but tune uses `.fit_pre()` and `.fit_model()` individually so it doesn't get run.

https://github.com/tidymodels/workflows/blob/40482ae456f3e96fc977dfa1bba60270a23e9f76/R/fit.R#L55-L75

I need to think a little about whether `.fit_model()` should just set `$trained <- TRUE` before it returns

DavisVaughan:
Okay, I have a lower level fix that will have to come in a few stages. Really workflows needs to export `.fit_finalize()` which would set `$trained` to `TRUE`, and then the code path functions like `fit_resamples()` would use it internally after calling the individual pieces of `.fit_pre()` and `.fit_model()` separately.

This will require a few preliminary PRs though, so I'm happy to handle this one

juliasilge:
OK, I'll take out the `trained` assignment and just leave in the tests, to be merged in after the changes in workflows and here. 👍

PR #304, opened 2020-10-19
Nest the grid-info inside the iteration function

DavisVaughan:
This should allow for alternate parallel processing strategies that require an unnested grid-info object in the foreach() loop

@topepo it seems that this was easier than expected 😛 . I now keep the grid-info object in its unnested form until we actually get into the iteration function. This should mean that we can do something like:

```r
# outer resample loop
foreach::foreach(
  iteration = seq_len(n_resamples),
  .packages = load_pkgs,
  .errorhandling = "pass"
) %:% {
  # inner flat loop over recipe/model combinations
  foreach::foreach(
    row = seq_len(nrow(grid_info)),
    .packages = load_pkgs,
    .errorhandling = "pass"
  ) %op% {
    grid_info <- vec_slice(grid_info, row)
    
    tune_grid_loop_iter_safely(
      iteration = iteration,
      resamples = resamples,
      grid_info = grid_info,
      workflow = workflow,
      metrics = metrics,
      control = control
    )
  }
}
```

### DIFF
--- a/R/grid_code_paths.R
+++ b/R/grid_code_paths.R
@@ -64,6 +64,26 @@ tune_grid_loop_iter <- function(iteration,
   model_param_names <- dplyr::pull(model_params, "id")
   preprocessor_param_names <- dplyr::pull(preprocessor_params, "id")
 
+  # Model related grid-info columns
+  cols <- rlang::expr(
+    c(
+      .iter_model,
+      .iter_config,
+      .msg_model,
+      tidyselect::all_of(model_param_names),
+      .submodels
+    )
+  )
+
+  # Nest grid_info:
+  # - Preprocessor info in the outer level
+  # - Model info in the inner level
+  if (tidyr_new_interface()) {
+    grid_info <- tidyr::nest(grid_info, data = !!cols)
+  } else {
+    grid_info <- tidyr::nest(grid_info, !!cols)
+  }
+
   split <- resamples$splits[[iteration]]
   training <- rsample::analysis(split)
 
--- a/R/grid_helpers.R
+++ b/R/grid_helpers.R
@@ -128,17 +128,20 @@ finalize_workflow_preprocessor <- function(workflow, grid_preprocessor) {
 # ------------------------------------------------------------------------------
 
 # For any type of tuning, and for fit-resamples, we generate a unified
-# grid-info object which is a tibble with two levels:
+# grid-info object which is a tibble with two layers of information:
 #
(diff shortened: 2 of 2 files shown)


topepo:
Not to jump the gun or anything but I tried the flattened loop (after importing `%:%`). The syntax can't have braces around the second `foreach()` call: 

```r
 results <- foreach::foreach(
    iteration = seq_len(n_resamples),
    .packages = load_pkgs,
    .errorhandling = "pass"
  ) %:%
    foreach::foreach(
      row = seq_len(nrow(grid_info)),
      .packages = load_pkgs,
      .errorhandling = "pass"
    ) %op% {
      grid_info <- vec_slice(grid_info, row)

      tune_grid_loop_iter_safely(
        iteration = iteration,
        resamples = resamples,
        grid_info = grid_info,
        workflow = workflow,
        metrics = metrics,
        control = control
      )
  }
```

The models come back with the error: 

```
"internal: Error in iter_grid_info_model[[\".submodels\"]][[1L]]: subscript out of bounds"
```

DavisVaughan:
its possible that that is overwriting the original grid-info. so maybe:

```r
 results <- foreach::foreach(
    iteration = seq_len(n_resamples),
    .packages = load_pkgs,
    .errorhandling = "pass"
  ) %:%
    foreach::foreach(
      row = seq_len(nrow(grid_info)),
      .packages = load_pkgs,
      .errorhandling = "pass"
    ) %op% {
      grid_info_row <- vec_slice(grid_info, row)

      tune_grid_loop_iter_safely(
        iteration = iteration,
        resamples = resamples,
        grid_info = grid_info_row,
        workflow = workflow,
        metrics = metrics,
        control = control
      )
  }
```

topepo:
That solved that issue but there is more to look at in the model loop. It fails with 

```
Error in iter_grid_info_model[[".iter_config"]][[1L]]: subscript out of bounds
``` 
On the second + iterations across models, this appears to be an empty list IFAICT. 

Here's what I was using to test: 

```r
library(tidymodels)

data(cells)

cells <- cells %>% dplyr::select(-case)
folds <- vfold_cv(cells)

xgb_spec <- boost_tree(trees = 50, min_n = tune()) %>% 
  set_engine("xgboost") %>% 
  set_mode("classification")

xgb_grid <- expand.grid(min_n = 2:10)

xgb_tune <- 
  xgb_spec %>% 
  tune_grid(class ~ ., folds, grid = xgb_grid)
```

which works well in the current PR.

PR #305, opened 2020-10-20
Adjust grid iteration and add hidden `parallel_method` control option

DavisVaughan:
Fixes the issues that appear when trying to use the alternate parallel method, as described in https://github.com/tidymodels/tune/pull/304#issuecomment-712867325.

Basically we need to use the `.iter_*` columns for iteration and filtering, not the number of rows.

I've also gone ahead and added the alternate parallel method through a hidden control option `parallel_method = c("outer", "flat")`. If not set it just defaults to the old `"outer"` method (I have no idea what to call this). But the example below will set it to the new "flat" mode

We can talk about what it should be called and make those changes here if you want. But this is at least a POC.

```r
library(tidymodels)
library(modeldata)

data(cells)

cells <- cells %>% dplyr::select(-case)
folds <- vfold_cv(cells)

xgb_spec <- boost_tree(trees = 50, min_n = tune()) %>% 
  set_engine("xgboost") %>% 
  set_mode("classification")

xgb_grid <- expand.grid(min_n = 2:10)

control <- control_grid()
control$parallel_method <- "flat"

xgb_tune <- 
  xgb_spec %>% 
  tune_grid(class ~ ., folds, grid = xgb_grid, control = control)
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: tune
 Title: Tidy Tuning Tools 
-Version: 0.1.1.9000
+Version: 0.1.1.9001
 Authors@R: c(
     person(given = "Max", family = "Kuhn", email = "max@rstudio.com", role = c("aut", "cre")),
     person("RStudio", role = "cph"))
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,8 @@
 
 ## Other Changes
 
+* New `parallel_over` control argument to adjust the parallel processing method that tune uses.
+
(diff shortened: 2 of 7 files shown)


DavisVaughan:
Possibly use `parallel_over = c("resamples", "everything")`

topepo:
I think default to "everything".

DavisVaughan:
That would be a change from the current behavior. I don't think we should do that because that would result in recipes being re-prepped many times since the loop is flattened

topepo (approved):

PR #314, opened 2020-10-30
move from crayon text coloring functions to cli

### LINKED ISSUE #311: replace crayon functions to those from cli
image

topepo:
Closes #311

The PR redefines crayon functions like `crayon::yellow()` to an internal function that uses `cli` via this pattern: 

```r
yellow <- function(...) as.character(cli::col_yellow(...))
```

The reason for the conversion is that some of the logging functions use character class objects (and `cli` gives us objects with class "ansi_string"). Perhaps @gaborcsardi  can tell me if this is the best way to go.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -23,7 +23,6 @@ Imports:
     ggplot2,
     glue,
     cli (>= 2.0.0),
-    crayon,
     yardstick (>= 0.0.7),
     rsample (>= 0.0.7.9000),
     tidyr,
--- a/R/logging.R
+++ b/R/logging.R
@@ -192,7 +192,7 @@ log_best <- function(control, iter, info, digits = 4) {
   }
 
   message("")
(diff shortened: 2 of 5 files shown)


gaborcsardi:
Maybe we'd just need an `as.character.ansi_string()` method?

DavisVaughan:
What actually happens if you don't do the forced conversion? Do you get an error?

topepo:
Yes, an error.

gaborcsardi:
Can you show some code that triggers the error?

gaborcsardi:
Reprex: 

```r
❯ siren("foobar", type = "warning")
Error: must have class `character`, not class `ansi_string`.
```

If I also add the `"character"` class then it is fine, so this is fixed here: https://github.com/r-lib/cli/commit/76fedd6d8e6d01da3eb193701b3f1f791ecbf723 but I think calling `as.character()` is perfectly fine as a workaround.

PR #315, opened 2020-10-30
Reexport tunable + required_pkgs from generics

### LINKED ISSUE #312: move to generics functions for required_pkgs and tunable
Once the current dev version is on CRAN

juliasilge:
Related to tidymodels/recipes#594
Closes #312 

This PR (using the development version of generics) solves our S3 struggles by reexporting `tunable()` and `required_pkgs()` from generics. We believe that this solution will (finally???) make parallel processing for Windows users work.

Open question 🤔 also mentioned on the recipes PR: there is a method for `tunable.step` in _both_ tune and recipes. Is that correct?


DavisVaughan:
See comments in https://github.com/tidymodels/recipes/pull/594 about not needing to lazy register

topepo:
Can you remove the remotes too?

topepo:
Can you also do the same for `min_grid()`?

juliasilge:
I **removed** the tunable method for `step` from tune and left it only in recipes FYI.

This looks like it is working well:

``` r
library(tidymodels)

pls_recipe <-
  step_pls(all_predictors(), outcome = "mpg")

required_pkgs(pls_recipe, FALSE)
#> [1] "mixOmics"

kpca_rec <- recipe(mpg ~ ., data = mtcars) %>%
  step_YeoJohnson(all_predictors()) %>%
  step_normalize(all_predictors()) %>%
  step_kpca_rbf(all_predictors())

required_pkgs(kpca_rec, FALSE)
#> [1] "dimRed"  "kernlab"
```

<sup>Created on 2020-11-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

R/min_grid.R
```diff
@@ -46,13 +46,6 @@
 #'   grid_regular(levels = 3)
 #'
 #' min_grid(xgb_spec, xgb_grid)
-#' @export
```
DavisVaughan, line 49: It might be good to have an explicit `NULL` after the last `#'` line.

I think this signals to roxygen that you are done with this free floating chunk of documentation

You'll probably need to add `@name min_grid` if you do that. I have a feeling that this chunk of documentation is matching up to the `min_grid.model_spec()` code below

R/load_ns.R
```diff
@@ -67,11 +67,6 @@ infra_pkgs <- c("tune", "recipes", "parsnip", "yardstick", "purrr", "dplyr",
 #' @param x An object.
 #' @return A character string.
 #' @keywords internal
-#' @export
-required_pkgs <- function(x, ...) {
```
DavisVaughan, line 71: See my note on the explicit `NULL` for `min_grid`

R/tunable.R
```diff
@@ -47,10 +47,6 @@
 #'   set_engine("C5.0", rules = TRUE) %>%
 #'   tunable()
 #' }
-#' @export
```
DavisVaughan, line 50: See my note about explicit `NULL` for `min_grid`

R/min_grid.R
DavisVaughan: Like this
https://github.com/DavisVaughan/almanac/blob/dc1e4d788fc8ab6557b95b76035437782e1841bd/R/adjustments.R#L49-L80

juliasilge: I think we _have_ to have the documentation match up to one of the methods. Otherwise, if we add a `@name`, it gives an error like `Rd files with duplicated alias 'min_grid'`. I think the reason it works for your almanac example is that you are not dealing with something reexported. Maybe a better thing to look at is [fit](https://github.com/tidymodels/parsnip/blob/master/R/fit.R) or [tidy](https://github.com/tidymodels/recipes/blob/master/R/tidy.R).

juliasilge:
There are some `tunable` methods for recipe steps from textrecipes in here, but I believe those belong in textrecipes. Maybe there was some trouble with registering `tunable` before; it should be easier now, just by reexporting.

topepo:
We'll move the `tunable` and `min_grid` functions to the corresponding packages but there will have to be some time in-between when they are in both.

juliasilge:
This is failing because tidymodels/recipes#594 is not merged in it yet; it does depend on that. The tests all pass for me locally.

R/min_grid.R
DavisVaughan: Okay thats fine. Can you move the block of pure comments starting with `As an example` up above the documentation? That way you can link the min-grid docs to the `min_grid.model_spec` method more clearly

Issue #300, opened 2020-10-11
last_fit() and fit() give different probability predictions for the same workflow and data


AndrewKostandy:
When the same workflow is used with last_fit() and fit() on the same training and testing data, the probability predictions are different. Below is a reproducible example where I fit the same workflow 4 times using an identical seed. Twice with last_fit() and twice with fit(). The 2 probability predictions of last_fit() are identical to each other and the 2 probability predictions of fit() are identical to each other, however the probability predictions of last_fit() and fit() are very different (not a rounding problem).

Note that this problem happens whether or not parallel processing is used - using the below line for example:
```r
# doParallel::registerDoParallel(cores = 7)
```
## Reproducible example

```r
library(tidyverse)
suppressMessages(library(tidymodels))
library(modeldata)
suppressMessages(library(themis))
library(embed)
data(attrition)

attrition <- janitor::clean_names(attrition)

attrition <- attrition %>% 
  mutate(attrition_class = fct_rev(attrition)) %>% 
  select(-attrition)

levels(attrition$attrition_class)
#> [1] "Yes" "No"

set.seed(1954)
attrition_split <- initial_split(attrition, strata = "attrition_class")
attrition_train <- training(attrition_split)
attrition_test <- testing(attrition_split)

xgboost_recipe <- 
  recipe(formula = attrition_class ~ ., data = attrition_train) %>% 
  step_upsample(attrition_class, seed = 936) %>% 
  step_discretize_xgb(age, daily_rate, distance_from_home, hourly_rate,
                      monthly_income, monthly_rate, total_working_years,
                      years_at_company, outcome = "attrition_class") %>% 
  step_novel(all_nominal(), -all_outcomes()) %>% 
  step_dummy(all_nominal(), -all_outcomes(), one_hot = TRUE) %>% 
  step_zv(all_predictors()) 

xgboost_spec <- 
  boost_tree(trees = 462, min_n = 9, tree_depth = 15, learn_rate = 0.02160385, 
             loss_reduction = 0.6299069, sample_size = 0.1487263) %>% 
  set_mode("classification") %>% 
  set_engine("xgboost") 

xgboost_workflow <- 
  workflow() %>% 
  add_recipe(xgboost_recipe) %>% 
  add_model(xgboost_spec)

common_seed <- 12345

# fit 2 models by last fit and 2 models by fit using the same seed
set.seed(common_seed)
last_fit1 <- last_fit(xgboost_workflow, attrition_split)

set.seed(common_seed)
[...]

PR #319, opened 2020-11-03
Ensure last_fit() uses RNG stream from main R process

topepo:
Fixes for #300 related to making sure that `last_fit()` and `fit.workflow()` produce the same results when a model uses random numbers. 

This PR also will change the `parallel_over` option form "resamples" to "everything" when a single resample is tested on multiple models (e.g. using a validation set). A message is issued when this occurs.

### DIFF
--- a/R/tune_grid.R
+++ b/R/tune_grid.R
@@ -380,7 +380,7 @@ tune_grid_workflow <- function(workflow,
   workflow <- set_workflow(workflow, control)
 
   if (any(names(resamples) == ".seed")) {
-    resamples <- resamples %>% dplyr::select(-.seed)
+    resamples <- dplyr::select(resamples, -.seed)
   }
 
   new_tune_results(
--- a/R/tune_grid.R
+++ b/R/tune_grid.R
@@ -347,13 +347,16 @@ tune_grid_workflow <- function(workflow,
 
   check_workflow(workflow, pset = pset)
 
-  resamples <- dplyr::mutate(resamples, .seed = sample.int(10^5, nrow(resamples)))
+  if (nrow(resamples) > 1 & !single_model(grid)) {
+    resamples <- dplyr::mutate(resamples, .seed = sample.int(10^5, nrow(resamples)))
+  }
 
   grid <- check_grid(
     grid = grid,
     workflow = workflow,
     pset = pset
   )
+
 
   # Save rset attributes, then fall back to a bare tibble
   rset_info <- pull_rset_attributes(resamples)
@@ -376,7 +379,9 @@ tune_grid_workflow <- function(workflow,
 
   workflow <- set_workflow(workflow, control)
 
-  resamples <- resamples %>% dplyr::select(-.seed)
+  if (any(names(resamples) == ".seed")) {
+    resamples <- resamples %>% dplyr::select(-.seed)
+  }
 
   new_tune_results(
     x = resamples,
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 
 * Fixed an issue where recipe tuning parameters could be randomly matched to the tuning grid incorrectly (#316).
 
-* `last_fit()` no longer accidentally adjusts the random seed (#264).
+* `last_fit()` no longer accidentally adjusts the random seed (#264) and will use sequential processing ()
 
 * Fixed two bugs in the acquisition function calculations.
 
@@ -34,7 +34,7 @@
 
 * `autoplot.tune_results()` now requires objects made by version 0.1.0 or higher of tune. 
 
-* `tune` objects no longer keep the `rset` class that they have from the `resamples` argument. 
+* tune objects no longer keep the `rset` class that they have from the `resamples` argument. 
 
 ## Other Changes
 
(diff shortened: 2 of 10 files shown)


R/tune_grid.R
```diff
@@ -376,7 +379,9 @@ tune_grid_workflow <- function(workflow,
 
   workflow <- set_workflow(workflow, control)
 
-  resamples <- resamples %>% dplyr::select(-.seed)
+  if (any(names(resamples) == ".seed")) {
+    resamples <- resamples %>% dplyr::select(-.seed)
```
juliasilge, line 383:
```suggestion
    resamples <- dplyr::select(resamples, -.seed)
```
Just to match the style for the other dplyr use here

juliasilge (commented):
I installed and used this (checked the extratests repo, along with other things, etc), and nothing surprising or bad cropped up.

I think I don't understand why we might want to force `parallel_over = "everything"` for a situation like a single validation set. I may not be understanding correctly, but if the preprocessing is slow, might not it be faster to use `parallel_over = "resamples"` even for a single resample, rather than process the preprocessor for each model?

topepo:
> I think I don't understand why we might want to force `parallel_over = "everything"` for a situation like a single validation set. I may not be understanding correctly, but if the preprocessing is slow, might not it be faster to use `parallel_over = "resamples"` even for a single resample, rather than process the preprocessor for each model?

Sure, if there is any severe preprocessing, this might be sub-optimal. However, with a single resample, `parallel_over = "resamples"`  doesn't do any parallel processing. It's pretty likely that any parallel processing is better than purely sequential processing.

juliasilge:
I am thinking about the example of text models, where the preprocessing often takes a lot longer than the model itself. Forcing the option here seems like it would make, for example, `last_fit()` take a lot longer than it would need to.

juliasilge:
I guess there's not much difference here, with the current main development branch installed and medium intense preprocessing and a pretty fast model:

``` r
library(tidymodels)
library(textrecipes)

data("small_fine_foods")
combined <- bind_rows(training_data, testing_data)
ind <- list(analysis = seq(nrow(training_data)), 
            assessment = nrow(training_data) + seq(nrow(testing_data)))
splits <- make_splits(ind, combined)
splits
#> <Analysis/Assess/Total>
#> <4000/1000/5000>

validation_set <- manual_rset(list(splits), ids = "validation")
validation_set
#> # Manual resampling 
#> # A tibble: 1 x 2
#>   splits          id        
#>   <list>          <chr>     
#> 1 <split [4K/1K]> validation

pre_proc <-
  recipe(score ~ review, data = training_data) %>%
  step_tokenize(
    review, token = "ngrams",
    options = list(
      n = 3, n_min = 1,
      stopwords = stopwords::stopwords(source = "snowball")
    )
  )  %>%
  step_tokenfilter(review, max_tokens = 4e3) %>%
  step_tfidf(review)

lasso_spec <-
  logistic_reg(penalty = 0.002, mixture = 1) %>%
  set_engine("glmnet")

sparse_bp <- hardhat::default_recipe_blueprint(composition = "dgCMatrix")

wf <- workflow() %>%
  add_recipe(pre_proc, blueprint = sparse_bp) %>%
  add_model(lasso_spec)

ctrl_everything <- control_resamples(parallel_over = "everything")
ctrl_resamples  <- control_resamples(parallel_over = "resamples")

doParallel::registerDoParallel()

res <- bench::mark(
  min_iterations = 10, check = FALSE, memory = FALSE,
  everything <- 
    fit_resamples(
      wf, 
      resamples = validation_set, 
      control = ctrl_everything
    ),
  resamples <-
    fit_resamples(
      wf, 
      resamples = validation_set, 
      control = ctrl_resamples
    )
)
#> 
#> Attaching package: 'rlang'
#> The following objects are masked from 'package:purrr':
#> 
#>     %@%, as_function, flatten, flatten_chr, flatten_dbl, flatten_int,
#>     flatten_lgl, flatten_raw, invoke, list_along, modify, prepend,
#>     splice
#> 
#> Attaching package: 'vctrs'
#> The following object is masked from 'package:tibble':
#> 
#>     data_frame
#> The following object is masked from 'package:dplyr':
#> 
#>     data_frame
#> Loading required package: Matrix
#> 
#> Attaching package: 'Matrix'
#> The following objects are masked from 'package:tidyr':
#> 
#>     expand, pack, unpack
#> Loaded glmnet 4.0-2
#> Warning: Some expressions had a GC in every iteration; so filtering is disabled.

res %>%
  select(-expression)
#> # A tibble: 2 x 5
#>        min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1    5.71s    6.08s     0.159        NA     1.29
#> 2    5.75s    5.93s     0.168        NA     1.07
```

<sup>Created on 2020-11-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

So maybe it won't make much difference in the end. I am a bit worried about turning this switch automatically and not letting people say what they want to do, though.

(Let me know if I did something wrong for benchmarking with parallel processing.)

PR #325, opened 2020-11-10
augment() methods for several object types from tune

topepo:


R/augment.R
```diff
@@ -0,0 +1,115 @@
+#' Augment data with holdout predictions
+#'
+#' For `tune` objects that use resampling, these `augment()` methods will add
+#' one or more columns for the hold-out predictions (i.e. from the assessment
+#' set(s)).
+#'
+#' @param x An object resulting from one of the `tune_*()` functions,
+#'  `fit_resamples()`, or `last_fit()`. The control specifications for these
+#'  objects should have used the option `save_pred = TRUE`.
```
juliasilge, line 9:
```suggestion
#' @param x An object resulting from one of the `tune_*()` functions,
#'  `fit_resamples()`, or `last_fit()`. The control specification for this
#'  object must have used the option `save_pred = TRUE` to generate
#'  predictions.
```

R/augment.R
```diff
@@ -0,0 +1,115 @@
...
+#' @param parameters A data frame with a single row that indicates what
+#' tuning parameters should be used to generate the predictions (for `tune_*()`
+#' objects only). If `NULL`, `select_best(x)` will be used.
+#' @param ... Not currently used.
+#' @return A data frame with one or more additional columns with model
+#' predictions.
```
juliasilge, line 15:
```suggestion
#' @return A data frame with one or more additional columns for model
#' predictions.
```

R/augment.R
```diff
@@ -0,0 +1,115 @@
...
+#'
+#' @details
+#' For some resampling methods where some rows are replicated in multiple
+#' assessment sets, the prediction columns will be averages of the holdout
+#' results. Also, for these methods, it is possible that all rows of the
+#' original data do not have holdout predictions. For example, a single
+#' bootstrap resample. In this case, all rows are return and a warning is
+#' issued.
```
juliasilge, line 23:
```suggestion
#' For some resampling methods where rows may be replicated in multiple
#' assessment sets, the prediction columns will be averages of the holdout
#' results. Also, for these methods, it is possible that all rows of the
#' original data do not have holdout predictions (like a single bootstrap
#' resample). In this case, all rows are return and a warning is issued.
```

juliasilge (approved):
This is really nice, and I'm glad we're adding it. 🙌 

Do you think an example would be good? I know examples + tests are already a little slow for this package overall. Another option would be to add it somewhere to `tidymodels.org`, like [maybe this article](https://www.tidymodels.org/learn/work/tune-svm/)?

topepo:
There are a few places in our course notes and examples that we can update. After the `tune` release, we should comb over everything to see what could be updated.

PR #326, opened 2020-11-11
adjust point size and alpha when racing

topepo:
Unit tests will be in the `finetune` package. 

Example with full grid search: 

image

Example with racing: 

image

### DIFF
--- a/R/plots.R
+++ b/R/plots.R
@@ -366,8 +366,8 @@ plot_marginals <- function(x, metric = NULL) {
 
   x <-
     x %>%
-    dplyr::rename(resamples = n) %>%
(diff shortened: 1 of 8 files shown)


R/plots.R
```diff
@@ -364,18 +366,30 @@ plot_marginals <- function(x, metric = NULL) {
 
   x <-
     x %>%
-    dplyr::select(dplyr::one_of(param_cols), mean, .metric) %>%
+    dplyr::rename(resamples = n) %>%
```
DavisVaughan, line 369: I think my only comment is that this might be _slightly_ clearer if this was `n_resamples`?

DavisVaughan (approved):

Issue #331, opened 2020-11-24
Parallel-safe RNG


HenrikBengtsson:
Hi, I think you want to use `doRNG::%dorng%` instead of just `%dopar%` to make sure you use the L'Ecuyer RNG in order to get statistically sound random numbers for the parallel processing.

You can detect this by running:

```r
library(doFuture)
registerDoFuture()
plan(multicore)  ## same as doMC()
example("tune_grid", package="tune")
```

which warns about this:

```r
> warnings()
Warning messages:
1: UNRELIABLE VALUE: Future ('doFuture-1') unexpectedly generated random numbers without specifying argument
'[future.]seed'. There is a risk that those random numbers are not statistically sound and the overall results
might be invalid. To fix this, specify argument '[future.]seed', e.g. 'seed=TRUE'. This ensures that proper, 
parallel-safe random numbers are produced via the L'Ecuyer-CMRG method. To disable this check,
use [future].seed=NULL, or set option 'future.rng.onMisuse' to "ignore".
...
```

When using `%dorng%`, the **doRNG** package will set up the L'Ecuyer RNG for you and these warnings go away.

PS. I'm working on making those RNG warnings more informative when using **foreach**, e.g. to have them refer to `doRNG::%dorng%` rather than the non-existing 'seed' argument

HenrikBengtsson:
PR #332 solves this;

```r
library(doFuture)
options(future.rng.onMisuse = "error") ## be strict; escalate RNG warnings to errors
registerDoFuture()
plan(multicore)  ## same as doMC()
example("tune_grid", package="tune")
```

FWIW, even if you don't use it, you need to keep import `%dopar%` due to the https://github.com/renozao/doRNG/issues/18 bug.

topepo:
The existing, albeit low-tech [solution](https://github.com/tidymodels/tune/blob/fc40e3033ff926be09ffbbec2192fbd0c69af650/R/grid_code_paths.R#L121), is to set a collection of reproduce seeds inside the functions called by the workers. Our tests show that this gives reproducible results across a bunch of backends (and sequentially). 

Are there cases where this would fail? 

I don't doubt that this is a more appropriate method for controlling randomness in the workers. 

[...]

PR #346, opened 2021-01-14
Reduce number of packages in Imports

juliasilge:
This PR reduces the number of packages in Imports down by three 3️⃣ :

- removing utils altogether
- moving tibble to Suggests (since dplyr imports it)
- moving hardhat to Suggests (since workflows imports it)

I think that is probably all we want to do for now. We need S3 methods from vctrs and an infix operator from rlang, so I believe it is better to keep those in Imports.

One point for discussion is GPfit, maybe for the future. The Bayesian methods are used much less often than the grid methods. If we get really hard up for "spots" in Imports, it might not be crazy unreasonable to move GPfit to Suggests and add `rlang::check_installed()` to guide folks to install it if they use `tune_bayes()`.

I did submit this PR to winbuilder and it passed fine. ✅


DavisVaughan:
I am fine with getting utils out of the way, I am a little uncomfortable about the other changes.

I would feel more comfortable about the hardhat -> Suggests move if we protect the 1 place that we directly call a hardhat function:
https://github.com/tidymodels/tune/blob/1f8f327ce7224ce29cb7ec62c4f150f36730622b/R/grid_helpers.R#L92

Maybe something like this right before that call:

```
if (!is_installed("hardhat") {
  abort("Internal error: hardhat should have been installed from the workflows dependency.")
}
```

I am a bit more uncomfortable with relying on dplyr to bring in tibble. I'm sure it will always continue to import tibble, but it still feels a bit wrong to me to do this since we use `tibble::` in so many places in tune. Can we get away with not moving this one to Suggests?

juliasilge:
We can probably get away without moving tibble to Suggests for now and address the RNG issues in #331 and #332 without that change, but it may be worth considering it in the future, depending on what comes up. Or we could consider GPfit maybe, if we have further problems.

juliasilge:
I'm happy with this PR after these last changes ☝️ because I think it gives up just enough space for now. We can revisit any more drastic changes and make tradeoffs in the future if we need to.

PR #349, opened 2021-02-01
Tweak tune random number seed generation and usage

### LINKED ISSUE #331: Parallel-safe RNG

DavisVaughan:
Closes #331 
Closes #332 

- Seeds are now generated with L'Ecuyer-CMRG, which is parallel safe.
- We now no longer repeatedly set the seed to the same value for each row of `rows` in the `"everything"` parallel loop

The strategy used here is slightly different from what we talked about in the group meeting. It was very difficult to generate and use 1 seed per row of `grid_info`. This was difficult to do because in `tune_grid_loop_iter()` we first loop over the preprocessors and fit them, then over the models. We _could_ just set the seed in that inner model loop, which is what we proposed in the group meeting. However, this would leave the preprocessor loop without a fixed seed, which I don't love because there could be randomness in some recipe steps that needs to be "fixed".

Because this was so difficult, I've just made a small tweak to the way we currently do things. Besides using the new type of random seeds, I've also just fixed the "bug" in the "everything" path where we were setting the seed to the same value repeatedly in the inner part of the nested loop. 

The way to explain our seed generation strategy now is: Every time `tune_grid_loop_iter()` is called, it sets a unique seed once at the top of the function call, then runs all the preprocessors and models in that particular iteration.

This means that the "everything" and "resamples" loop still don't use identical random seed strategies, but I'm honestly not sure how to fix that without a complete rewrite of the tune paths.

### DIFF
--- a/R/control.R
+++ b/R/control.R
@@ -127,6 +127,11 @@ control_resamples <- control_grid
 #'   If `NULL`, chooses `"resamples"` if there are more than one resample,
 #'   otherwise chooses `"everything"` to attempt to maximize core utilization.
 #'
+#'   Note that switching between `parallel_over` strategies is not guaranteed
+#'   to use the same random number generation schemes. However, re-tuning a
+#'   model using the same `parallel_over` strategy is guaranteed to be
+#'   reproducible between runs.
+#'
 #' @details
 #'
 #' For `extract`, this function can be used to output the model object, the
--- a/R/grid_code_paths.R
+++ b/R/grid_code_paths.R
@@ -17,16 +17,13 @@ tune_grid_loop <- function(resamples, grid, workflow, metrics, control, rng) {
   parallel_over <- control$parallel_over
   parallel_over <- parallel_over_finalize(parallel_over, n_resamples)
 
-  if (rng) {
-    seeds <- sample.int(10^5, nrow(resamples))
-  } else {
-    seeds <- NULL
-  }
-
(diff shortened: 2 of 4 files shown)


DavisVaughan:
Note: Will add docs about the fact that the "everything" and "resamples" loop can use different RNG strategies

PR #353, opened 2021-02-17
pillar methods for tune objects

topepo:
Will require new versions of `pillar` and `tibble` to fully work so `size_sum()` is conditionally registered.


DESCRIPTION
```diff
@@ -46,5 +46,4 @@ LazyData: true
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.1.9000
 Language: en-US
-Remotes:
-    tidymodels/rsample
```
DavisVaughan, line 50: Did you want to require a specific cran rsample version now? i.e. i dont think it should be `rsample (>= 0.0.8.9001)` anymore, but maybe `rsample (>= 0.0.9)`?

R/0_imports.R
```diff
@@ -39,7 +41,7 @@ utils::globalVariables(
     ".extracts", ".metrics", "value", ".notes", ".loss", ".bound",
     ".column", ".totals", ".value", "direction", ".config", "Freq", "Prediction",
     "Truth", ".seed", ".order", ".iter_model", ".iter_preprocessor",
-    ".iter_config", ".msg_model", "# resamples")
+    ".iter_config", ".msg_model", "# resamples", "seed")
```
DavisVaughan, line 44: How did you make the global variable warning appear? I never saw it

R/pillar.R
```diff
@@ -0,0 +1,45 @@
...
+
+#' Helpers for pillar formatting
+#' @param x an object
+#' @return A character string.
+#' @export
+#' @keywords internal
+#' @rdname pillar-helpers
+size_sum.tune_results <- function(x) {
+  ""
+}
```
DavisVaughan, line 21: I am uncomfortable with `@export`ing this

Since you don't do `@importFrom tibble size_sum`, this will export an actual function named `size_sum.tune_results()`. i.e. you will be able to do `tune::size_sum.tune_results()` which we definitely do not want.

I think it you remove the export tag (and all the other documentation) and re-document, then it should still work as expected since we lazy register it in the onLoad instead

To make it clear that this function will be registered in the onLoad, i typically add a comment (not a roxygen comment!) like:

```
# @export - lazy in .onLoad()
```

as done in:
https://github.com/r-lib/clock/blob/56cf09a260e5b9328e18a8fa24f77221191143a8/R/calendar.R#L32

R/pillar.R
```diff
@@ -0,0 +1,45 @@
...
+    res <- "rsmp[+]"
+  } else {
+    res <- "rsmp[x]"
+  }
+  res
+}
+
+#' @export
+#' @keywords internal
+#' @rdname pillar-helpers
```
DavisVaughan, line 37: Same here, no need to export

R/pillar.R
DavisVaughan: It is possible that you will also need to change the name from `size_sum.tune_results` to `size_sum_tune_results`. In rsample, I had to name methods this way when they were being conditionally registered based on a package version https://github.com/tidymodels/rsample/blob/9caeda19ce8fa603d6a42827f8a685b80f573c68/R/compat-dplyr-old.R#L22

It has something to do with the fact that `devtools::load_all()` will always try and register them, since they "look like" S3 methods based on the `.` in their name

R/zzz.R
```diff
@@ -79,6 +79,13 @@
     vctrs::s3_register("dplyr::dplyr_reconstruct", "resample_results", method = dplyr_reconstruct_resample_results)
     vctrs::s3_register("dplyr::dplyr_reconstruct", "iteration_results", method = dplyr_reconstruct_iteration_results)
   }
+
+  if (newer_tibble()) {
+    vctrs::s3_register("tibble::obj_sum", "tune_results")
+    vctrs::s3_register("tibble::obj_sum", "resample_results")
```
DavisVaughan, line 85: You don't need to conditionally register these two. They will automatically be registered since you `#' @export` them and did `@importFrom tibble obj_sum`

R/zzz.R
```diff
@@ -79,6 +79,13 @@
     vctrs::s3_register("dplyr::dplyr_reconstruct", "resample_results", method = dplyr_reconstruct_resample_results)
     vctrs::s3_register("dplyr::dplyr_reconstruct", "iteration_results", method = dplyr_reconstruct_iteration_results)
   }
+
+  if (newer_tibble()) {
```
DavisVaughan, line 83: You may want to add a `# TODO` here that says that when the new version of tibble goes to CRAN, we should:

- bump the required tibble version
- remove these conditional registrations
- register methods with the traditional `@importFrom tibble size_sum` and `@export` system

DESCRIPTION
topepo: I left it as-is to the CI runs won't fail for the next week (for binaries to be built). I can change it later.

R/0_imports.R
topepo: Local R CMD check with R version 4.0.3 (2020-10-10). That was the motivation for my `tune` question for you the other day. ¯\_(ツ)_/¯

R/zzz.R
topepo: It looks like Kirill is releasing `tibble` in the next few days so we might not need this at all. I'll wait to merge this until that is on CRAN (assuming that it is on CRAN by the 2021-03-01 deadline for `tune`).

PR #356, opened 2021-02-26
Fix issue with finalize_recipe when individual recipe steps have multiple parameters to be tuned

stevenpawley:
### The Problem

Hello, I came across a small issue where `tune::finalize_recipe` fails if a step within a recipe object has multiple parameters that have been tuned. I originally had this problem on a custom recipe, but I've found a step function in recipes that also reproduces it:

``` r
library(tidymodels)
library(modeldata)
data(biomass)

model_spec <- linear_reg() %>%
  set_engine("lm")

rec_obj <- 
  recipe(HHV ~ carbon + hydrogen + oxygen + nitrogen + sulfur, data = biomass) %>%
  step_isomap(all_predictors(), neighbors = tune(), num_terms = tune())

params <- rec_obj %>%
  parameters() %>%
  update(
    neighbors = dials::neighbors(c(10, 100)),
    num_terms = num_terms(c(2, 4))
  ) %>%
  grid_max_entropy(size = 10)

resamples <- bootstraps(biomass, times = 1)

tune_res <-
  tune_grid(
    model_spec,
    preprocessor = rec_obj,
    grid = params,
    metrics = metric_set(rsq),
    resamples = resamples
  )

best <- select_best(tune_res, "rsq")
finalize_recipe(rec_obj, best)
#> Error in names(param) <- pset$name: 'names' attribute [2] must be the same length as the vector [1]
```

<sup>Created on 2021-02-25 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

### Proposed Solution

The problem occurs in the `tune:::complete_steps` function, as well as in the `tune::finalize_recipe`

```
complete_steps <- function (param, pset, object) 
{
    step_ids <- purrr::map_chr(object$steps, ~.x$id)
    step_index <- which(pset$component_id == step_ids)
    tmp <- object$steps[[step_index]]
    list_para <- list(param)
    names(param) <- pset$name
    tmp <- rlang::exec(update, object = tmp, !!!param)
    object$steps[[step_index]] <- tmp
    object
}
```

The `step_index <- which(pset$component_id == step_ids)` assumes that the parameter set has a single component id, but a recipe step with multiple parameters will have the same component id (for the step) for each parameter.

This also required a modification to `finalize_recipe`:

```
  for (i in seq_along(pset)) {
    pset_params <- parameters[names(parameters) %in% pset[[i]]$id]. # new addition
    x <- complete_steps(pset_params, pset[[i]], x)
  }
```

Where the parameters for the particular pset are extracted, and then passed to complete steps.

The modification passes the test cases and appears to work for recipes with single or multiple steps, and steps that include a custom parameter id, i.e. `tune(id = "my_nearest_neighbors")`. See if you think that this resolves the issue and doesn't have any other (negative) effects.


topepo (changes requested):
Good catch. 

The changes look good. 

Can you add a unit test that uses steps from within `recipes` (instead of `embed`)? 

Also, please add a note to the NEWS file.

PR #390, opened 2021-06-29
changes for RNG consistency in #389

### LINKED ISSUE #389: more RNG consistency
First, we should make sure that package dependencies do not [change the RNNG state](https://stackoverflow.com/questions/49640909/why-would-an-r-package-load-random-numbers) when they are loaded. For example: 

```
> set.seed(1)
> runif(1)
[1] 0.2655087
> 
> set.seed(1)
> library(forecast)
Registered S3 method overwritten by 'quantmod':
  method            from
  as.zoo.data.frame zoo 
> runif(1)
[1] 0.3721239
```

Second, currently `tune` resets the RNGkind after usage: 

``` r
library(tidymodels)
#> Registered S3 method overwritten by 'tune':
#>   method                   from   
#>   required_pkgs.model_spec parsnip

RNGkind()
#> [1] "Mersenne-Twister" "Inversion"        "Rejection"
```

``` r
set.seed(6735)
```

``` r
spline_rec <-
  step_ns(disp, deg_free = tune("disp")) %>%
  step_ns(wt, deg_free = tune("wt"))

lin_mod <-
  linear_reg() %>%
  set_engine("lm")

spline_grid <- expand.grid(disp = 2:5, wt = 2:5)

spline_res <-
  tune_grid(lin_mod, spline_rec, resamples = folds, grid = spline_grid)

RNGkind()
#> [1] "L'Ecuyer-CMRG" "Inversion"     "Rejection"
```

<sup>Created on 2021-06-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

We do change the RNG kind (see #331) and [reset it](https://github.com/tidymodels/tune/blob/87a79e4dd01483aa97f02d4d95993360d1421850/R/grid_code_paths.R#L470:L473).  However, it is being reset somewhere so we need to add another `on.exit()`.

topepo:
`load_namespace()` was altered to reset the RNG state in case packages use random numbers on startup. 

We also added an additional `on.exit()` to `tune_grid_loop()` so that RNGkind stays consistent. 

closes #389

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -35,6 +35,7 @@ Imports:
     tibble (>= 3.1.0),
     tidyr,
     vctrs (>= 0.3.0),
+    withr,
     workflows (>= 0.2.2),
     yardstick (>= 0.0.7)
 Suggests: 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # tune (development version)
 
+* When using `load_pkgs()`, packages that use random numbers on start-up do not affect the state of the RNG. We also added more control of the RNGkind to make it consistent with the user's previous value (#389). 
+
 # tune 0.1.5
 
 * Fixed a bug where the resampled confusion matrix is transposed when `conf_mat_resamped(tidy = FALSE)` (#372)
--- a/R/grid_code_paths.R
+++ b/R/grid_code_paths.R
@@ -2,6 +2,12 @@ tune_grid_loop <- function(resamples, grid, workflow, metrics, control, rng) {
   `%op%` <- get_operator(control$allow_par, workflow)
   `%:%` <- foreach::`%:%`
 
+  current_rng_kind <- RNGkind()[[1]]
+  on.exit(
+    RNGkind(kind = current_rng_kind),
+    add = TRUE
+  )
+
   tune_grid_loop_iter_safely <- super_safely_iterate(tune_grid_loop_iter)
 
   packages <- c(control$pkgs, required_pkgs(workflow))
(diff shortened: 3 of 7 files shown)


DavisVaughan:
Okay I figured out what is changing the random seed "kind". `generate_seeds()` is doing the right thing, but our `assign()` statement inside the tune loop is what borks things:

Specifically, this line here:
https://github.com/tidymodels/tune/blob/f48f2323d967a3610b0537ecd9cc47abceef8658/R/grid_code_paths.R#L128

Which I would advise changing to this, and hopefully that will fix it:

```r
# `assign()`-ing the random seed alters the `kind` type to L'Ecuyer-CMRG,
# so we have to ensure it is restored on exit
old_kind <- RNGkind()[[1]]
assign(".Random.seed", seed, envir = globalenv())
on.exit(RNGkind(kind = old_kind), add = TRUE)
```

Here is a minimal reprex of what is happening:

``` r
# Not here...
RNGkind()
#> [1] "Mersenne-Twister" "Inversion"        "Rejection"
seeds <- tune:::generate_seeds(rng = TRUE, n = 5)
RNGkind()
#> [1] "Mersenne-Twister" "Inversion"        "Rejection"

# But this changes it
assign(".Random.seed", seeds[[1]], envir = globalenv())
RNGkind()
#> [1] "L'Ecuyer-CMRG" "Inversion"     "Rejection"
```

DavisVaughan:
Regarding the loading of packages and how they might change the RNG stream - we are already very careful to load known packages _before_ we set the random seed, so this shouldn't be an issue? I guess it probably doesn't hurt to add it to `load_pkgs()`, but I don't think it fixes any current issues with `tune_grid()`.
https://github.com/tidymodels/tune/blob/f48f2323d967a3610b0537ecd9cc47abceef8658/R/grid_code_paths.R#L123-L129

PR #405, opened 2021-09-08
use new dials objects for xgboost engine args

### LINKED ISSUE #402: Add `lambda` and `alpha` as tunable args for boosted trees with the xgboost engine
related: https://github.com/tidymodels/dials/issues/176

add them to `tunable.boost_tree()`

hfrick:
Dials now contains new parameter object for xgboost-specific engine args so they are added to the result of `tunable()` here. 

Closes #402 for `alpha` and `lambda` - and I've added `scale_pos_weight` as well.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -54,3 +54,5 @@ Language: en-US
 LazyData: true
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.1.9001
+Remotes:
+  tidymodels/dials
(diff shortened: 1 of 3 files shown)


DESCRIPTION
```diff
@@ -54,3 +54,5 @@ Language: en-US
 LazyData: true
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.1.9001
+Remotes:
+  tidymodels/dials
```
topepo, line 58: Do we need a to have `dials (>= 0.1.10)`?

topepo (changes requested):

PR #406, opened 2021-09-08
drop usage of deprecated `step_knnimpute()`

hfrick:
This PR switches the use of  the deprecated `step_knnimpute()` to the new `step_impute_knn()` and ups the required version of recipes


juliasilge (approved):
Looks great! 🙌

PR #410, opened 2021-09-15
Merge steps using a S3 merge method

dfalbel:
Implement merge for recipes steps as a S3 method so package developers can override the behavior for specific steps.
I think it makes sense to have this since `merge` is defined as: 'Merge parameter grid values into objects' and that seems to be what we were doing with:

```
object$steps[[ idx ]][[param_info$name]] <- grid[[i]]
```

For an example, we implement a `merge` specific method here to be able to delegate from the meta step to the actual step:

https://github.com/dfalbel/maybestep/blob/main/R/maybestep.R#L108


topepo:
Thanks but I'll close this; I don't think that we'll use this potential feature.

PR #416, opened 2021-10-01
remove `tune_args()` generic

hfrick:
This is a joint PR with https://github.com/r-lib/generics/pull/60 to move the `tune_args()` generic from tune to generics so that we can add methods for various objects across different tidymodels packages more easily.


juliasilge (changes requested):
This should probably go into [`reexports.R`](https://github.com/tidymodels/tune/blob/master/R/reexports.R) like the other things from generics, right?

Also, I think we probably don't want to merge this until the generics PR is merged, because having users install from branches gets a little messy (having too many people install from branches when they may not understand that is what they are doing might get to be a mess). I suggest waiting until the generics PR is done.

R/tune_args.R
```diff
@@ -1,60 +1,9 @@
...
-#' boost_tree(trees = tune(), min_n = tune()) %>%
-#'   set_engine("xgboost") %>%
-#'   tune_args()
-#'
-#' boost_tree(trees = tune(), min_n = tune()) %>%
-#'   set_engine("C5.0", rules = TRUE) %>%
-#'   tune_args()
-#' }
+#' @name tune_args
 #' @export
```
DavisVaughan, line 5: I don't think you even need a `@name` or `@keywords` tag here. We don't want roxygen docs generated for this. After all the movement has been done, I think tune is just going to re-export `tune_args()` and won't provide any of the methods (so there is nothing to provide docs for)

R/tune_args.R
```diff
@@ -1,60 +1,3 @@
...
-#'   set_engine("C5.0", rules = TRUE) %>%
-#'   tune_args()
-#' }
-#' @export
-tune_args <- function(object, ...) {
-  UseMethod("tune_args")
-}
 
 #' @export
 #' @rdname tune_args
```
juliasilge, line 3: This whole file should just go away, right, like Davis mentioned? We only will reexport it and there will be nothing here at all?

juliasilge (changes requested):

R/tune_args.R
hfrick: That is the overall plan, yes. It is still part of this PR because I wanted to first do a joint PR on generics and tune to move the generic itself and then do another joint PR to move the methods because that involves tune and recipes/parsnip/workflows.

hfrick: Will leave such context in the PR description next time!

PR #418, opened 2021-10-04
Notes improvements and collect_notes()

topepo:


NEWS.md
```diff
@@ -1,5 +1,9 @@
 # tune (development version)
 
+* The `.notes` column now contains information on the type of note (error or warning), the location where it occurred, and the note. Printing a tune result has different output describing the notes. 
+
+* `collect_notes()` can be used to gather any notes to a tibble.
```
EmilHvitfeldt, line 5:
```suggestion
* `collect_notes()` can be used to gather any notes to a tibble. (#363)
```

EmilHvitfeldt (approved):
I like it! General question, what is the reason behind the ordering of the columns returned by `collect_notes()`? I was expecting the `note` column to come first. Actually, I was expecting the columns in the complete opposite order.

topepo:
I think that most people see any messaging and think that it is an error (based on SO or GH questions). Plus the note is the longest field and might obscure the others when the tibble is printed.

PR #425, opened 2021-11-03
Find SVG files for article

juliasilge:
If you look at the pkgdown site for tune:

- [release](https://tune.tidymodels.org/articles/acquisition_functions.html)
- [dev](https://tune.tidymodels.org/dev/articles/acquisition_functions.html)

you'll notice that we have broken links to images on the acquisition functions article. I poked around in the pkgdown documentation, tried a few things, and discovered that setting these [as external resources](https://pkgdown.r-lib.org/dev/reference/build_articles.html#external-files) solves the problem.

### DIFF
--- a/vignettes/acquisition_functions.Rmd
+++ b/vignettes/acquisition_functions.Rmd
@@ -6,6 +6,11 @@ vignette: >
 output:
   knitr:::html_vignette:
     toc: yes
+resource_files:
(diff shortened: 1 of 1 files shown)


juliasilge:
Working [now](https://tune.tidymodels.org/dev/articles/acquisition_functions.html)

PR #437, opened 2021-11-23
add note about logging when running in parallel

### LINKED ISSUE #435: `control_grid(verbose = TRUE)` does not produce more output
Shouldn't this print some logging results while tuning? 

``` r
library(tidymodels)
data(cells)

set.seed(1697)
folds <- vfold_cv(cells, v = 3)

xgb_spec <- boost_tree(trees = 500, min_n = tune()) %>%
  set_engine('xgboost') %>%
  set_mode('classification')

cells_wf <- workflow(class ~ ., xgb_spec)

doParallel::registerDoParallel()
set.seed(12)
tune_grid(
  cells_wf,
  resamples = folds,
  control = control_grid(verbose = TRUE)
)
#> # Tuning results
#> # 3-fold cross-validation 
#> # A tibble: 3 × 4
#>   splits             id    .metrics          .notes          
#>   <list>             <chr> <list>            <list>          
#> 1 <split [1346/673]> Fold1 <tibble [20 × 5]> <tibble [0 × 3]>
#> 2 <split [1346/673]> Fold2 <tibble [20 × 5]> <tibble [0 × 3]>
#> 3 <split [1346/673]> Fold3 <tibble [20 × 5]> <tibble [0 × 3]>
```

<sup>Created on 2021-11-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

hfrick:
Closes #435

### DIFF
--- a/R/control.R
+++ b/R/control.R
@@ -1,6 +1,13 @@
 #' Control aspects of the grid search process
 #'
 #' @inheritParams control_bayes
+#' @param verbose A logical for logging results as they are generated. Despite
+#'   this argument, warnings and errors are always shown. When running in
+#'   parallel, almost all of the logging will not be seen. This is dependent on
+#'   the backend package and technology being used. If using a dark IDE
+#'   theme, some logging messages might be hard to see. If this is the case,
+#'   try setting the `tidymodels.dark` option with
+#'   `options(tidymodels.dark = TRUE)` to print lighter colors.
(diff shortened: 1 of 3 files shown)


juliasilge:
Instead of doing it here, could we do it in `control_bayes()` so we can `inheritParams`? The same issue applies to `control_bayes()`.

hfrick:
ah, if it applies to both then sure!

Issue #419, opened 2021-10-08
tune logo: green or grey text?


hfrick:
The [logo in the repo](https://github.com/tidymodels/tune/blob/master/man/figures/logo.png) has grey text and is from December 2019. 

The [logo in the hex-sticker repo](https://github.com/rstudio/hex-stickers/blob/master/PNG/tune.png) has green text and is from April 2020.

Should the logo here be updated to the green version?

juliasilge:
Yes, I believe we got it updated but never updated it here in the repo. 🙈 It should be the one in the `/rstudio/hex-stickers`.

PR #445, opened 2022-01-22
update hex logo

EmilHvitfeldt:
This PR aims to close https://github.com/tidymodels/tune/issues/419.


juliasilge:
Can you tell me more about why the new PNG here in the PR is smaller than what we had before? The new one is 240 px or so. I know it's not the best for web performance (for pkgdown, etc) to have a bigger PNG but the full-size versions we typically put in a repo are not _that_ big (~100 KB). I do think it would be good to keep at least a larger version in the package repo itself (if not full size), rather than having to go to a different repo to find a version with decent resolution.

EmilHvitfeldt:
I thought it was weird too, but that is what happened when I used `use_logo("~/Github/hex-stickers/PNG/tune.png")`. I have manually copied over the file in full resolution

juliasilge (approved):
Based on the discussion we had today, let's plan for the full size versions to live in https://github.com/rstudio/hex-stickers and just keep small versions in the package itself. That sounds good to me!

PR #446, opened 2022-02-04
deprecate `parameters()` methods

hfrick:
This PR soft-deprecates the `parameters()` methods for recipe, model_spec, and workflow objects which have been superseded by the corresponding `extract_parameter_set_dials()` methods. The new methods are in the corresponding packages (i.e. in recipe, parsnip, and workflows).


R/param_set.workflows.R
```diff
@@ -1,5 +1,10 @@
 #' Determination of parameter sets for other objects
 #'
+#' @description
+#' `r lifecycle::badge("deprecated")`
+#'
+#' These methods have been deprecated in favor of `extract_parameter_set_dials()`.
```
DavisVaughan, line 6:
```suggestion
#' These methods have been deprecated in favor of [extract_parameter_set_dials()].
```

and then redocument

R/param_set.workflows.R
```diff
@@ -37,96 +42,70 @@
 #'   set_engine("C5.0", rules = TRUE) %>%
 #'   dials::parameters()
 #' }
+#'
+#' # ->
```
DavisVaughan, line 46: I don't think you need the examples, I think redirecting is enough.

tests/testthat/test-param_set.R
```diff
@@ -7,18 +7,24 @@ source(test_path("../helper-objects.R"))
 # ------------------------------------------------------------------------------
 
 test_that('recipe with no steps', {
+  withr::local_options(lifecycle_verbosity = "quiet")
+
   bare_info <- dials::parameters(bare_rec)
   check_param_set_tibble(bare_info)
   expect_equal(nrow(bare_info), 0)
```
DavisVaughan, line 14: Since the methods are already tested in their respective packages, I would remove all of these tests.

If you'd like, you could keep 1 simple test per method in this package, just to ensure that calling `parameters()` "still works" for now - but it should be clear in the testthat message that it is deprecated

Something like:
https://github.com/r-lib/vctrs/blob/main/tests/testthat/test-lifecycle-deprecated.R

`local_lifecycle_silence()` is just https://github.com/r-lib/vctrs/blob/e86bb9a4b57459eb56099c0b52af37877a9bf515/R/compat-lifecycle.R#L149

DavisVaughan (changes requested):

tests/testthat/test-param_set.R
hfrick: Regarding`local_lifecycle_silence()`: do you mean we should use something with `rlang::local_options()` here instead of a function from `withr`? Or is the second link just for illustration?

DavisVaughan: just for illustration. you can use either withr or rlang here for `local_options()`, they are basically equivalent

hfrick:
Failures are due to 
* new class of errors from dplyr --> addressed in  #447
* randomForest requiring R > 4.1.0 --> should we replace `rand_forest()` in the tests with `svm_linear()` since we also have kernlab in Suggests? Alternatively, we could move the tests to extratests.

DavisVaughan (approved):

PR #448, opened 2022-02-11
Drop randomForest

hfrick:
randomForest now requires R => 4.1.0 so all checks on `oldrel-1` and older fail.

Where possible I swapped out `rand_forest(engine = "randomForest")` for `svm_linear(engine = "kernlab")` because `kernlab` is already a dependency for tune.

One test requires tuning a parameter which per default has an `unknown()` value. The two main parameters for `svm_linear()` don't have that. I've currently moved the test to extratests https://github.com/tidymodels/extratests/pull/42. The alternative could be to take on a different dependency for an engine package where we have a parameter with an `unknown()` default, in order to keep them in tune.

Candidate parameters and their respective dependencies would be
- `sample_size()` which is used in `boost_tree()` with the engines `xgboost` and `c5`
- `batch_size()` which is an engine arg to `mlp()` with engines `keras` and `nnet`
- `num_comp()` and `num_terms()` which I think gets used in recipes (`step_pca()`)
- `mtry()` which is used in `rand_forest()` with the engine `ranger` 

Does any of these sound like a good choice? or rather keep it in extratests?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -42,7 +42,6 @@ Suggests:
     kernlab,
     knitr,
     modeldata,
-    randomForest,
     spelling,
     testthat,
     xml2
--- a/tests/testthat/test-augment.R
+++ b/tests/testthat/test-augment.R
@@ -7,7 +7,6 @@ library(dials)
 data(two_class_dat, package = "modeldata")
 
 lr_spec <- logistic_reg() %>% set_engine("glm")
-rf_spec <- rand_forest(trees = 20, min_n = tune()) %>% set_engine("randomForest") %>% set_mode("regression")
 
 set.seed(1)
 bt1 <- bootstraps(two_class_dat, times = 30)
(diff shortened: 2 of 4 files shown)


DavisVaughan (approved):
I think out of all of those options, C50 might be a good choice because it is fairly straightforward to install, it has a low R version dependency, it is already Suggested by parsnip, and (most importantly) we control it.

So you could add a C50 Suggests dep just for this one test I guess. It may come in handy later.

I also don't mind if this test lives in extratests if that proves to be problematic for any reason, but it does seem like a test that _should_ live in tune if it is easy to do.

hfrick:
I started with switching to C5.0 but found out that while `dials::sample_size()` does have `unknown()` in the default and `boost_tree()` has a main argument of the same name, it actually uses `sample_prop()` for which defaults don't depend on the data. 

I think the choice boils down to 
1) exchange `randomForest` for `ranger`, i.e., give `tune` a Suggests dependency on `ranger` or
2) move the test to `extratests`

Any preference?

hfrick:
I don't think it's a bug: The [engine docs for C5.0](https://parsnip.tidymodels.org/reference/details_boost_tree_C5.0.html#arguments) do say that it's the proportion. I assume the main arg of `boost_tree()` is called `sample_size` because the default for `xgboost` is the number, not the proportion.

Issue #449, opened 2022-02-18
Uninformative error if wrong `control_*()` is used


EmilHvitfeldt:
See reprex below. This should be a fairly easy since since the control functions return specific classes we can check against.

``` r
library(tidymodels)

rec_no_tune_1 <-
  recipe(mpg ~ ., data = mtcars) %>%
  step_normalize(all_predictors())

set.seed(363)
mt_folds <- vfold_cv(mtcars, v = 5)

glmn_mod <- decision_tree(tree_depth = tune(), min_n = 2) %>% 
  set_mode("regression") %>%
  set_engine("rpart")

workflow() %>%
  add_recipe(rec_no_tune_1) %>%
  add_model(glmn_mod) %>%
  tune_bayes(mt_folds, control = control_grid())
#> Error in if (is.na(limit)) {: argument is of length zero
```

<sup>Created on 2022-02-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

topepo:
In iterative tuning methods and racing, we pass a single control function for the function (e.g. `tune_sim_anneal()`) and that is used to work with another function, usually `tune_grid()`. 

I think that we need a set of `as_control_grid()` methods that will convert from one control type to another. We’d need these to go from more complete methods to less complex. For example, we would need `as_control_resamples.control_grid()` but not `as_control_grid.control_resamples()`.

PR #452, opened 2022-02-19
Give informative error when wrong control_*() function is used

EmilHvitfeldt:
This PR attempts to close https://github.com/tidymodels/tune/issues/449.

Locally I'm getting this weird error, which for the life of me I cannot figure out why

```
── Failure (test-bayes.R:440:3): missing performance values ────────────────────
`expect_error(...)` did not produce any messages.
```


R/checks.R
```diff
@@ -533,3 +533,11 @@ check_gp_failure <- function(current, prev) {
   # return prev model
   prev
 }
+
+check_control <- function(control, name) {
+  if (!inherits(control, name)) {
+    rlang::abort(
```
simonpcouch, line 539: Since this is a user-facing error, I would set `call = NULL` in `rlang::abort` so that the user isn't shown:

```
Error in `check_control()` 
```

prepending the error.

tests/testthat/test-bayes.R
```diff
@@ -447,6 +447,19 @@ test_that("too few starting values", {
   expect_snapshot(error = TRUE, tune:::check_bayes_initial_size(1, 1, FALSE))
 })
 
+test_that("error if wrong control function is used", {
+  set.seed(4400)
+  wflow <- workflow() %>% add_recipe(rec_tune_1) %>% add_model(lm_mod)
+  pset <- dials::parameters(wflow) %>% update(num_comp = dials::num_comp(c(1, 5)))
+  folds <- vfold_cv(mtcars)
```
simonpcouch, line 454: Could you namespace calls to rsample/dials/parsnip in these tests?

simonpcouch (changes requested):
I think this is a super helpful change that's worth making!

A question I don't know the answer to:

How far can users usually get before this becomes an issue? If sometimes quite far, do we expect this to be a breaking change for many existing users?

If this isn't just a matter of failing more informatively a bit earlier on—i.e. if folks can sometimes get to their desired result while getting this wrong—we could transition this to an `rlang::warn` (though this would still require the same amount of legwork for hard revdeps on CRAN).

tests/testthat/test-bayes.R
EmilHvitfeldt: done

EmilHvitfeldt:
> How far can users usually get before this becomes an issue? If sometimes quite far, do we expect this to be a breaking change for many existing users?

As far as I can tell, this just provides better error messages.

simonpcouch (approved):
Thanks for making those edits! Looks much better to me.

I didn't realize this the first time around, but `control_resamples` and `control_grid` are _aliases_, and thus do not error when interchanged. I think this is the right behavior.

So, to clarify:

``` r
library(tune)
library(waldo)

res <- control_resamples()

grid <- control_grid()

bayes <- control_bayes()

compare(res, grid)
#> ✔ No differences

compare(res, bayes)
#> `old` is length 8
#> `new` is length 12
#> 
#> `class(old)`: "control_grid"  "control_resamples"
#> `class(new)`: "control_bayes"                    
#> 
#>     names(old)      | names(new)                      
#> [1] "verbose"       | "verbose"         [1]           
#> [2] "allow_par"     - "no_improve"      [2]           
#>                     - "uncertain"       [3]           
#>                     - "seed"            [4]           
#> [3] "extract"       | "extract"         [5]           
#> [4] "save_pred"     | "save_pred"       [6]           
#>                     - "time_limit"      [7]           
#> [5] "pkgs"          | "pkgs"            [8]           
#> [6] "save_workflow" | "save_workflow"   [9]           
#>                     - "save_gp_scoring" [10]          
#> ... ...               ...               and 2 more ...
#> 
#> `old$allow_par` is a logical vector (TRUE)
#> `new$allow_par` is absent
#> 
#> `old$no_improve` is absent
#> `new$no_improve` is an integer vector (10)
#> 
#> `old$uncertain` is absent
#> `new$uncertain` is a double vector (Inf)
#> 
#> `old$seed` is absent
#> `new$seed` is an integer vector (64749)
#> 
#> `old$time_limit` is absent
#> `new$time_limit` is a logical vector (NA)
#> 
#> `old$save_gp_scoring` is absent
#> `new$save_gp_scoring` is a logical vector (FALSE)

inherits(res, "control_resamples")
#> [1] TRUE
inherits(res, "control_grid")
#> [1] TRUE
inherits(res, "control_bayes")
#> [1] FALSE
```

<sup>Created on 2022-05-26 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Adding a second revision to `NEWS` (sorry) to reflect this, as this wasn't clear to me until reading the `control_*` source. LGTM once NEWS is clarified. :)

A note for @topepo's review:

This does not introduce issues for stacks, as `control_stack_*` functions wrap the tune analogues directly and don't add their own class. Do you anticipate any issues here with `finetune` or any other packages that extend tune?

EmilHvitfeldt:
This will be closed in favor of https://github.com/tidymodels/tune/pull/540

Issue #431, opened 2021-11-11
Controlling metric event_level for last_fit()


mattwarkentin:
## The problem

Hi,

Maybe I'm missing something, but how does `last_fit.workflow` know which `event_level` to use for metric computations? All other fitting functions allow the user to specify this via `control` argument, but this function defines the controller inside the function itself (see below) via `control_resamples()`. `control_resamples()` has `event_level = 'first'` by default, yet when I use `last_fit()` it returns the metrics for `event_level = 'second'`, which is actually what I want. I am just not sure where `last_fit` is inheriting this behaviour.

https://github.com/tidymodels/tune/blob/6897071ff1201e333692db890b87bbc909a46706/R/last_fit.R#L103-L108

## Reprex

``` r
library(tidymodels)

data <- mutate(mtcars, vs = factor(vs))

tt_split <- initial_split(data, prop = 8/10, strata = 'vs')
rs <- bootstraps(training(tt_split), times = 1, strata = 'vs')

wflow <-
  workflow() %>%
  add_model(
    logistic_reg(penalty = tune(), mixture = tune()) %>%
      set_engine('glmnet')
  ) %>%
  add_formula(vs ~ .)

results <- tune_grid(
  wflow,
  resamples = rs,
  grid = 1,
  control = control_grid(event_level = 'second')
)

show_best(results, metric = 'roc_auc')
#> # A tibble: 1 × 8
#>    penalty mixture .metric .estimator  mean     n std_err .config             
#>      <dbl>   <dbl> <chr>   <chr>      <dbl> <int>   <dbl> <chr>               
#> 1 0.000227   0.749 roc_auc binary      0.95     1      NA Preprocessor1_Model1

finalize_workflow(
  wflow,
  select_best(results, metric = 'roc_auc')
) %>%
  last_fit(tt_split) %>%
  collect_metrics()
#> # A tibble: 2 × 4
#>   .metric  .estimator .estimate .config             
#>   <chr>    <chr>          <dbl> <chr>               
#> 1 accuracy binary             1 Preprocessor1_Model1
#> 2 roc_auc  binary             1 Preprocessor1_Model1
```

juliasilge:
I believe this is the same as #399 but let me know if I am mistaken

mattwarkentin:
Indeed this looks like a duplicate issue. On the plus side, I'm glad I wasn't missing something obvious; on the down side, my searching skills could be better.

PR #481, opened 2022-04-12
Update last_fit to allow the passing of a control object

### LINKED ISSUE #399: control_last_fit()
The other `tune` functions have control arguments/functions. That's where we set options like `event_level`. 

See [this SO post](https://stackoverflow.com/questions/68561374/logistic-regression-with-tidymodels-how-to-set-event-level-second-in-last-fi/68616042).

mattwarkentin:
Fixes #399 (and related to #431)


R/control.R
```diff
@@ -70,6 +70,41 @@ print.control_grid <- function(x, ...) {
...
+#'
+#' @details
+#'
+#' [control_last_fit()] is a wrapper around [control_resamples()] and is meant
+#'   to be used with [last_fit()].
+#'
+#' @export
+control_last_fit <- function(
+    verbose = FALSE,
+    pkgs = NULL,
```
juliasilge, line 85: Can you say more about why you included `pkgs` as an argument? Is this used for anything other than parallel processing? The other args that control parallel processing are not included, for good reason.

tests/testthat/_snaps/bayes.md
```diff
@@ -57,38 +61,41 @@
...
 
 ---
 
     Code
       tune:::check_bayes_initial_size(5, 1, FALSE)
-    Error <rlang_error>
-      ! There are 5 tuning parameters and a single grid point was requested. The GP model requires 2+ initial points but there should be more initial points than there are tuning paramters. 
+    Condition
+      Error in `tune:::check_bayes_initial_size()`:
+      ! ! There are 5 tuning parameters and a single grid point was requested. The GP model requires 2+ initial points but there should be more initial points than there are tuning paramters.
```
juliasilge, line 82: Just want to highlight how the errors look now @topepo; we probably want to address this.

juliasilge:
Looks like we may have a problem with oldrel-3 now, due to the new kernlab release?

R/control.R
mattwarkentin: No, I think you are right. The `pkgs` argument can probably be removed. Want me to push this change?

mattwarkentin: Fixed with [cb41113](https://github.com/tidymodels/tune/pull/481/commits/cb41113a27ddccfa60890bc61a47e6b437a822b6)

tests/testthat/_snaps/bayes.md
topepo: Do you mean the function called back? I'd leave as-is for this PR and we can add an issue to improve it overall.

juliasilge:
> Looks like we may have a problem with oldrel-3 now, due to the new kernlab release?

We asked some experts and it looks like kernlab needs R 3.6.2, because they are missing `FCONE` on older versions. I took `oldrel-3` out, but we can likely add it back in after the next major version bump in R.

juliasilge (approved):
Thank you so much @mattwarkentin! 🙌

PR #493, opened 2022-05-23
add comments suggesting to debug `tune_grid_loop_iter`

simonpcouch:
As I often guess wrong on the first try.😄

### DIFF
--- a/R/grid_code_paths.R
+++ b/R/grid_code_paths.R
@@ -35,6 +35,7 @@ tune_grid_loop <- function(resamples, grid, workflow, metrics, control, rng) {
           ns = "tune"
         )
 
+        # Likely want to debug at `tune_grid_loop_iter`
         tune_grid_loop_iter_safely(
           split = split,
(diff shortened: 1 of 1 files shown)


R/grid_code_paths.R
```diff
@@ -35,6 +35,7 @@ tune_grid_loop <- function(resamples, grid, workflow, metrics, control, rng) {
           ns = "tune"
         )
 
+        # Likely want to debug at `tune_grid_loop_iter`
```
DavisVaughan, line 38:
```suggestion
        # Likely want to debug with `debugonce(tune_grid_loop_iter)`
```

How about this, which contains the exact thing you can copy and paste?

DavisVaughan (approved):
LGTM, feel free to merge after addressing that one comment and letting CI pass

R/grid_code_paths.R
simonpcouch: You got it. :)

PR #497, opened 2022-05-26
Use agua's iteration function for h2o engine in `tune_grid_loop`

qiushiyan:
This PR lets `tune_grid_loop` uses agua's `tune_grid_loop_iter_h2o` when model uses h2o engine. We made `make_safely_iter` to return the correct safely iter function that will be passed to foreach.


R/grid_code_paths.R
```diff
@@ -5,13 +5,9 @@ tune_grid_loop <- function(resamples, grid, workflow, metrics, control, rng) {
...
 
 
-  is_h2o <- is_h2o(workflow)
   # TODO: remove when h2o supports non-regular grid
-  regular_grid <- is.null(grid) || is_regular_grid(grid)
-  if (is_h2o) {
-    if (!regular_grid) {
-      rlang::abort("The `h2o` engine only supports regular tuning grid.")
-    }
+  if (!is.null(grid) && is_h2o(workflow) && !is_regular_grid(grid)) {
```
qiushiyan, line 9: h2o currently only supports regular grid. If the grid is null then we are in `fit_resamples`. no need to check.

R/grid_code_paths.R
```diff
@@ -443,12 +460,16 @@ is_failure <- function(x) {
...
 }
 
-parallel_over_finalize <- function(parallel_over, n_resamples) {
+parallel_over_finalize <- function(parallel_over, n_resamples, is_h2o = FALSE) {
   # Always use user supplied option, even if not as efficient
   if (!is.null(parallel_over)) {
     return(parallel_over)
   }
 
+  if (is_h2o) {
```
qiushiyan, line 469: hard code parallel_over to 'resamples' for h2o engine to match h2o.grid

PR #505, opened 2022-06-03
Further abstract `tune_grid_loop()` to support agua

DavisVaughan:
Closes #497

I think #497 is awesome. I wanted to see if I could push it a little further to confine all of the agua specific details into one place.  That way we have a way to more easily replicate this in the future if we have to do this again. It would have been fairly hard to convey these details through PR review, so I've done a new PR here. I think, I'm fairly happy with what I've come up with - but it wouldn't have been possible without @qiushiyan's proof of concept in #497!

### DIFF
--- a/R/grid_code_paths.R
+++ b/R/grid_code_paths.R
@@ -1,26 +1,161 @@
-tune_grid_loop <- function(resamples, grid, workflow, metrics, control, rng) {
-  `%op%` <- get_operator(control$allow_par, workflow)
-  `%:%` <- foreach::`%:%`
+tune_grid_loop <- function(resamples,
+                           grid,
+                           workflow,
+                           metrics,
+                           control,
+                           rng) {
+  fn_tune_grid_loop <- tune_grid_loop_tune
 
-  packages <- c(control$pkgs, required_pkgs(workflow))
+  if (workflow_uses_agua(workflow)) {
+    # Special allowance for agua, which uses a custom loop implementation
+    fn_tune_grid_loop <- tune_grid_loop_agua
+  }
 
-  grid_info <- compute_grid_info(workflow, grid)
+  results <- fn_tune_grid_loop(
+    resamples,
+    grid,
+    workflow,
+    metrics,
+    control,
+    rng
+  )
 
+  resamples <- pull_metrics(resamples, results, control)
+  resamples <- pull_notes(resamples, results, control)
+  resamples <- pull_extracts(resamples, results, control)
+  resamples <- pull_predictions(resamples, results, control)
+  resamples <- pull_all_outcome_names(resamples, results)
+
+  resamples
+}
+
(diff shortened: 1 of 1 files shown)


R/grid_code_paths.R
```diff
@@ -349,16 +475,17 @@ tune_grid_loop_iter <- function(split,
 
 # ------------------------------------------------------------------------------
 
-tune_grid_loop_iter_safely <- function(split,
+tune_grid_loop_iter_safely <- function(fn_tune_grid_loop_iter,
```
qiushiyan, line 478: is there a way to split what this function does into separate functions
- make the safely version of the passed iter function and return it
- gather grid results
- postprocess errors, notes, etc.

R/grid_code_paths.R
```diff
@@ -1,26 +1,161 @@
...
+  packages <- c(control$pkgs, required_pkgs(workflow))
+  grid_info <- compute_grid_info(workflow, grid)
+
+  n_splits <- length(splits)
+  n_grid_info <- nrow(grid_info)
+
+  # Must assign `tune:::tune_grid_loop_iter_safely()` to a function in
+  # this function environment so foreach can find it. Mainly an issue with
+  # doParallel and PSOCK clusters.
+  fn_tune_grid_loop_iter_safely <- tune_grid_loop_iter_safely
```
qiushiyan, line 150: if returning a function, could use something like
`fn_tune_grid_loop_iter <- make_safely(fn_tune_grid_loop_iter)`, and then use `fn_tune_grid_loop_iter` in foreach

R/grid_code_paths.R
DavisVaughan: Okay, so I took another look at this.

I don't think this is actually going to work, mainly because of line 517 which is:

```r
notes <- log_problems(notes, control, split, "internal", problems)
```

We log problems about the iteration results _as they occur_.

If we were to move the post processing step out of the foreach loop, then we'd be forced to log all of the iteration result problems _after_ all of the iterations have run. This would probably result in messages that look out of order since `fn_tune_grid_loop_iter()` itself can also log problems.

i.e. we might get a message on iteration 3 from inside `fn_tune_grid_loop_iter()`, but if iteration 2 fails then the message for that (handled by the safely wrapper) would show up _after_ the message about iteration 3 if we move the post processing out of the foreach loop.

Since we can't cleanly separate the pre/iter/post calls into 3 pieces, I feel like it is probably best to just keep them together as is for now

DavisVaughan:
@qiushiyan could you please test this PR in combination with agua locally on your own computer? I just want to make sure that it is working as you expect (since we can't add tests here)

qiushiyan:
> @qiushiyan could you please test this PR in combination with agua locally on your own computer? I just want to make sure that it is working as you expect (since we can't add tests here)

It worked and tuning tests are passing.

PR #506, opened 2022-06-04
better error message for fit_resamples and last_fit with tuning parameters 

topepo:
``` r
library(tidymodels)
tidymodels_prefer()
theme_set(theme_bw())

rec <- recipe(mpg ~ ., data = mtcars) %>% step_ns(disp, deg_free = tune())
spec <- linear_reg(penalty = tune()) %>% set_engine("glmnet")
wflow <- workflow(rec, spec)

rs <- vfold_cv(mtcars)

fit_resamples(wflow, rs)
#> Error in `check_no_tuning()` at tune/R/resample.R:124:2:
#> ! 2 arguments have been tagged for tuning in these components: model_spec and recipe. 
#> Please use one of the tuning functions (e.g. `tune_grid()`) to optimize them.
```

<sup>Created on 2022-06-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


R/checks.R
```diff
@@ -533,3 +533,19 @@ check_gp_failure <- function(current, prev) {
...
+  if (num_param == 0) {
+    return(invisible(FALSE))
+  }
+  srcs <- unique(tune_param$source)
+  num_srcs <- length(srcs)
+  srcs <- paste0(srcs, collapse = " and ")
+  msg_1 <- cli::pluralize("{num_param} argument{?s} {?has/have} been tagged for tuning in {?this component/these components}: {srcs}. ")
+  msg_2 <- "Please use one of the tuning functions (e.g. `tune_grid()`) to optimize them."
+  msg <- paste(msg_1, msg_2, sep = "\n")
+  rlang::abort(msg)
```
simonpcouch, line 549:
```suggestion
  rlang::abort(msg, call = NULL)
```

simonpcouch: Since this is a user-facing error, we can hide the source line/helper name here:

``` r
boop <- function() {
  rlang::abort("woah!")
}

bop <- function() {
  rlang::abort("woah!", call = NULL)
}

boop()
#> Error in `boop()`:
#> ! woah!

bop()
#> Error:
#> ! woah!
```

<sup>Created on 2022-06-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

It looks like snapshots may not need to be updated with this change.

simonpcouch (approved):
I love this! Should be really helpful.

PR looks super clean. Thumbs up from me, with one suggestion.

R/checks.R
juliasilge: One thing we might want to think about with error messages is to move away from bespoke `\n` because the rlang / cli infrastructure handles that better relative to the console window. Instead of pasting with `\n`, we can use a vector that handles it automatically:

``` r
rlang::abort(c("Oh no!", "This is a real problem"))
#> Error:
#> ! Oh no!
#> • This is a real problem
```

<sup>Created on 2022-06-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

topepo: This is being fixed in #513 where we get all of the lines:


> res_roles <- role_bp_wflow %>% fit_resamples(rs)
x Fold01: preprocessor 1/1, model 1/1 (predictions):
  Error:
  ! The following cols are missing from `new_data`: date.
  ℹ There are also non-standard recipe roles for the column(s).
  ℹ See `?update_role` for more information on how use non-standard recipe roles during prediction.
x Fold02: preprocessor 1/1, model 1/1 (predictions):
  Error:
  ! The following cols are missing from `new_data`: date.
  ℹ There are also non-standard recipe roles for the column(s).
  ℹ See `?update_role` for more information on how use non-standard recipe roles during prediction.
```

and so on.

PR #508, opened 2022-06-06
update machinery for bayesian tuning prompting

### LINKED ISSUE #483: Fix error messaging for `tune_bayes()`
The [snapshot tests for `tune_bayes()`](https://github.com/tidymodels/tune/blob/main/tests/testthat/_snaps/bayes.md) now include some ‼️ BANG BANG that I don't believe we want around:

```
! ! There are 5 tuning parameters and a single grid point was requested. The GP model requires 2+ initial points but there should be more initial points than there are tuning paramters. 
```

It also looks pretty weird interactively.

simonpcouch:
Closes #483. :)

Transitions internals to `cli` tooling and addresses the duplicate `! !` prefix.🐨

### DIFF
--- a/R/checks.R
+++ b/R/checks.R
@@ -207,8 +207,10 @@ check_bayes_initial_size <- function(num_param, num_grid, race = FALSE) {
     rlang::abort(
       c(
         msg,
-        glue::glue("The GP model requires 2+ initial points but there should \\
-                    be more initial points than there are tuning parameters."),
+        glue::glue(
+          "The GP model requires 2+ initial points. For best performance, \\
+          supply more initial points than there are tuning parameters."
+        ),
         race_msg
       ),
       call = NULL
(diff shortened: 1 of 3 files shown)


juliasilge (approved):
This looks great! I just have a question about the existing message:

> The GP model requires 2+ initial points but there should be more initial points than there are tuning parameters.

This may be me not understanding the GP model super well, but I find the "but" confusing here, like I am expecting the message to tell me what I did wrong. Should this be "and", like:

> The GP model requires at least 2 initial points, and more initial points than there are tuning parameters.

simonpcouch:
I agree with you. :)

I think these checks are split up between `abort()` (at `num_grid == 1`) and `inform()` (at `num_grid < num_param + 1` because the former is a requirement and the latter is a suggestion, so I think we should still phrase the second as a suggestion.

Maybe:

> The GP model requires 2+ initial points. For optimum performance, supply more initial points than there are tuning parameters (i.e. at least {num_param}).


The later message should also indicate _why_ it's messaging. i.e. 

> This is likely to cause numerical issues in the first few search iterations.

to

> `diff <- num_param - num_grid`

> There are {qty(diff)}{?as many/more} tuning parameters ({num_param}) {?as/than} there are initial points ({num_grid}). This is likely to cause numerical issues in the first few search iterations.

Issue #500, opened 2022-05-28
PR: Some best practices and new tests


franzbischoff:
## Pull Request

First, let me apologize for the previous PR. That was after a night long and in a rush.
Now I've read the contribution docs and I hope I can be useful.

On this [PR](https://github.com/tidymodels/tune/pull/499)

Suggestions:

- `dplyr::one_of` was superseded by `dplyr::all_of`
- Added namespaces on several external packages usage
- Removed unnecessary namespaces on test which causes incompatibility with VSCode Testing suits and no harm with RStudio.
- Added #nolint statements where `lintr` complains about not using `snake_case` format.
- Added 2 unit tests for `multi_predict`

franzbischoff:
Related to https://github.com/tidymodels/tune/issues/463

franzbischoff:
Hello,

As discussed on last closed PR, it was agreed that:

- `dplyr::one_of` may be changed by `dplyr::all_of` in a new PR.
- Some namespaces can be removed on test units, in a new PR.

1. Tests for multi_predict can be added? Should be a separate PR or the same as the PR for namespaces above?
2. Removing the call to `library(ggplot2)` on `test-autoplot.R`, an error arises that may have been hidden, can this be fixed in the same PR for test units?
```r
object 'CoordFixed' not found
ggplot2::ggproto_parent(CoordFixed, self)
       at tune/R/coord_obs_pred.R:16:6
```

Thank you.

EmilHvitfeldt:
1. and 2. I would put those into separate PRs as well. 2. especially since it likely will need to change some of the .R files

simonpcouch:
Looks like the linked PRs here have been addressed. Thanks!

PR #512, opened 2022-06-11
Test units for multi_predict

franzbischoff:
Related to Issue https://github.com/tidymodels/tune/issues/500

This PR adds two unit tests for the `multi_predict` function.


juliasilge:
Can you give us a bit more background for your thoughts on adding these tests here? The tests for `multi_predict` and its methods [are in parsnip](https://github.com/tidymodels/parsnip/search?q=multi_predict).

franzbischoff:
Well, that is an example of looking "too near" from the tree. 😄 
Since I was focusing on the `tune` package I didn't realize that `parsnip` had that covered.
What brought me to adding such test in `tune` however was the fact that there is a `test-predict.R` here on `tune` and I missed the `multi_predict` since it also can be used for _tuning_ purposes.

I'll review this in `parsnip` and I'll get back here later. Maybe the _tuning_ process using `multi_predict` is the only suitable part here.

Thanks.

juliasilge:
Can you tell me a bit more about what you want to test here that is not already [tested in extratests](https://github.com/tidymodels/extratests/blob/main/tests/testthat/test-engine-parameters-c5.R)? We've got multiple trees already in those tests.

franzbischoff:
Right! Well, I think this solves this PR.
Sorry to make you dig into this!
I'll be more careful next time.

Best regards!

PR #513, opened 2022-06-14
added show_notes

topepo:


R/tune_grid.R
```diff
@@ -341,7 +341,7 @@ tune_grid_workflow <- function(workflow,
   )
 
   if (is_cataclysmic(resamples)) {
-    rlang::warn("All models failed. See the `.notes` column.")
+    rlang::warn("All models failed. Run `show_notes(your_object_name)` for more information.")
```
juliasilge, line 344:
```suggestion
    rlang::warn("All models failed. Call `show_notes()` on your results for more information.")
```
It's not super clear what `your_object_name` is in this context.

topepo: Draft PR. I'd wait before adding comments

topepo: We're going to stash the `tune_results` objects at the end of their respective functions so that we can say `show_notes(.Last.tune.results)`

PR #531, opened 2022-08-02
Add backend options in control functions 

qiushiyan:
related tidymodels/agua#29


topepo:
At first pass, I think that `backend_options` should be an argument for the control function(s)

R/tune_grid.R
```diff
@@ -249,7 +250,8 @@ tune_grid.default <- function(object, ...) {
 #' @rdname tune_grid
 tune_grid.model_spec <- function(object, preprocessor, resamples, ...,
                                  param_info = NULL, grid = 10, metrics = NULL,
-                                 control = control_grid()) {
+                                 control = control_grid(),
+                                 backend_options = list()) {
```
DavisVaughan, line 254: Oh I have an idea to make this even more self documenting. Rather than `list()`, I think you could allow `backend_options` to be one of two things

```
- Defaults to `NULL`, for default backend options

- A `<tune_backend_options>` object, as created by `new_backend_options()`.
```

tune would export `new_backend_options()`, and then agua would use that to create and export its own `agua_backend_options()` function. Like this:

```r
library(rlang)

# Exported from tune for use in agua
new_backend_options <- function(..., class = character()) {
  out <- list2(...)
  
  if (any(names2(out) == "")) {
    abort("All backend options must be named.")
  }
  
  structure(out, class = c(class, "tune_backend_options"))
}

# Then in agua you'd export and document: 
agua_backend_options <- function(parallelism = 1L) {
  new_backend_options(parallelism = parallelism, class = "agua_backend_options")
}

# In the future we could add this directly in tune, if the tune backend
# ever had options that were specific to it. But right now we don't have
# any, so we don't need this function.
backend_options <- function(tune_backend_option = NULL) {
  new_backend_options(tune_backend_option = tune_backend_option)
}
```

tune would be in charge of checking that `backend_options` is either `NULL` or `inherits(backend_options, "tune_backend_options")` and then would pass it on to the backend.

agua would check that `backend_options` is either `NULL` or `inherits(backend_options, "agua_backend_options")`.

Does that make sense?

DavisVaughan (changes requested):

R/tune_grid.R
qiushiyan: Thats looks great to me and I will make a patch. I have no opinon on if `backend_options` should be put in control functions, are we set on this?

topepo: Yes, put them in the control functions (incl the ones in fine-tune).

qiushiyan:
todo: add validation

PR #537, opened 2022-09-07
Racing s3 methods

topepo:
We should only report or select the model configurations that ran the full race (i.e. were fully resampled). 

This PR has new S3 methods for `tune_race` objects that work appropriately.


R/collect.R
```diff
@@ -349,16 +376,33 @@ collect_metrics.tune_results <- function(x, summarize = TRUE, ...) {
...
   }
+  if (!complete) {
+    final_configs <- race_subset(x)
+    res <- dplyr::inner_join(res, final_configs, by = ".config")
+  }
   res
 }
 
+# Only return configurations that were completely resampled
+race_subset <- function(x) {
```
hfrick, line 395: A name like `subset_finished_race()` or similar could capture the comment above about which subset exactly

inst/test_objects.R
```diff
@@ -1,4 +1,9 @@
+# ------------------------------------------------------------------------------
+# Run this, in place, from the terminal with R CMD BATCH --vanilla test_objects.R
+# Make sure to commit the Rout file
```
hfrick, line 3: This currently clashes with the bullet point for the release questions which suggests running R CMD BATCH in the pkg root directory (which is why it had all the paths constructed with `testthat::test_path()`. I don't have strong opinions on which way we do it, but it should be consistent :D 

https://github.com/tidymodels/tune/blob/a86603c0f3ad39b59d88c11977560a20cc78a759/R/0_imports.R#L57

hfrick (commented):
A general question first: Is there a particular reason that you want to add these methods in tune rather than finetune? The latter would be in keeping with the principle of putting methods in the same package as the class.

When reading through this, I had to pay extra attention to how the word "complete" is used. The description of the `complete` arg in the `@param` section talks about the "complete set of results", i.e. all configurations. The `@details` section then talks about the configurations that were "completely resampled" - which is also close to "configs that completed the race". I think it would be easier to understand if we use "complete" only for one of those two (opposing) options. I like "configurations which finished the race" and "complete set of configurations".

topepo:
Moved to tidymodels/finetune#49

PR #541, opened 2022-09-15
update snapshot with new version of pillar

EmilHvitfeldt:
This PR updates snapshots using {pillar} 1.8.1 so we don't get `# i Use `print(n = ...)` to see more rows`


simonpcouch:
It looks like the merge upstream brought in these same changes—fair game to close?

EmilHvitfeldt:
Sure thing!

PR #546, opened 2022-09-23
add missing collect.md snaps

EmilHvitfeldt:
It appears to me that https://github.com/tidymodels/tune/pull/544 accidentally removed `_snaps/collect.md`. This PR should fix it


simonpcouch (approved):
I see the same output after `devtools::test()`, thumbs up from me!

PR #547, opened 2022-09-23
Add verbose_progress argument to control_bayes()

EmilHvitfeldt:
Related to https://github.com/tidymodels/finetune/issues/44.

### DIFF
--- a/R/tune_bayes.R
+++ b/R/tune_bayes.R
@@ -258,7 +258,7 @@ tune_bayes_workflow <-
 
     score_card <- initial_info(mean_stats, metrics_name, maximize)
 
-    if (control$verbose_progress) {
+    if (control$verbose_iter) {
       message_wrap(paste("Optimizing", metrics_name, "using", objective$label))
     }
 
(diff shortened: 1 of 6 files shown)


R/tune_bayes.R
```diff
@@ -618,12 +618,7 @@ more_results <- function(object, resamples, candidates, metrics, control, param_
         param_info = param_info,
         grid = candidates,
         metrics = metrics,
-        control = control_grid(
-          verbose = FALSE,
-          extract = control$extract,
-          save_pred = control$save_pred,
-          event_level = control$event_level
-        )
+        control = parsnip::condense_control(control, control_grid())
```
EmilHvitfeldt, line 621: This will pass more argument through then previously, but it seems like the right thing to do

EmilHvitfeldt: I realized that we don't even need this as `tune_grid()` does this under the hood

EmilHvitfeldt:
@topepo once this PR is approved, then I'll make the corresponding finetune PR

topepo:
Can you add some testing snapshots for a small problem?

PR #549, opened 2022-09-26
add prompt for params that will not be tuned

### LINKED ISSUE #548: No error/warning for mismatch between model arguments and set_engine()
I ran into the following error today. And it took me an embarrassingly long time to figure out that the error happened because there was a mismatch between engine and model arguments. I debated where to put this issue since it could both be in {tune} or {parsnip}.
Maybe it would be worth it to do a check in `set_engine()` to make sure that the supplied arguments are useable.

``` r
library(tidymodels)

set.seed(6735)

car_rec <-
  step_normalize(all_predictors())

lasso_spec <-
  linear_reg(penalty = tune(), mixture = 1) %>%
  set_mode("regression") %>%
  set_engine("lm")

# Use a space-filling design with 7 points
set.seed(3254)
svm_res <- tune_grid(lasso_spec, car_rec, resamples = folds, grid = 7)
#> Warning: No tuning parameters have been detected, performance will be evaluated
#> using the resamples with no tuning. Did you want to [tune()] parameters?
```

<sup>Created on 2021-10-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

simonpcouch:
Closes #548.

I came across some similar confusion to Emil's in his issue #548. That issue was originally on parsnip, but I think implementing this check in tune makes more sense, as it isn't relevant outside of parameter tuning and we'd run into the same challenges as we do with other parsnip model spec checking, where it's not immediately clear when a user is "done" defining a model specification.

[EDIT: this PR originally made changes to both the existing warning and added a new one. now, it only adds a new error.]

``` r
library(tidymodels)
```

The error flags when a user marks parameters for tuning when they can’t be tuned due to choice of engine.

``` r
tune_grid(
  linear_reg(engine = "lm", penalty = tune()),
  mpg ~ .,
  bootstraps(mtcars, 2)
)
#> Error in `tune_grid()`:
#> ! The parameter `penalty` was marked with `tune()`, though will not be
#>   tuned.
#> ℹ This usually means that the current modeling engine `lm` does not support
#>   tuning `penalty`.

tune_bayes(
  linear_reg(engine = "lm", penalty = tune(), mixture = tune()),
  recipe(mpg ~ ., mtcars) %>% step_ns(hp, deg_free = tune()),
  bootstraps(mtcars, 2)
)
#> Error in `tune_bayes()`:
#> ! The parameters `penalty` and `mixture` were marked with `tune()`,
#>   though will not be tuned.
#> ℹ This usually means that the current modeling engine `lm` does not support
#>   tuning `penalty` and `mixture`.
```

Note that this PR does not change either the conditions or content of the warning shown in the original issue. That warning is raised to address a different issue, and this error catches the cases that that warning erroneously was made to handle. :)

``` r
tune_grid(
  linear_reg(),
  mpg ~ .,
  bootstraps(mtcars, 2)
)
#> Warning: No tuning parameters have been detected, performance will be evaluated
#> using the resamples with no tuning. Did you want to [tune()] parameters?
#> # Tuning results
#> # Bootstrap sampling 
#> # A tibble: 2 × 4
#>   splits          id         .metrics         .notes          
#>   <list>          <chr>      <list>           <list>          
#> 1 <split [32/11]> Bootstrap1 <tibble [2 × 4]> <tibble [0 × 3]>
#> 2 <split [32/13]> Bootstrap2 <tibble [2 × 4]> <tibble [0 × 3]>
```

See test cases for confirmation that we don’t raise it when we ought not to, incl. compatibility with recipes and bayesian tuning. Comment below addresses why this is an error rather than a warning.

<sup>Created on 2022-09-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


simonpcouch:
This should also help us immediately surface bugs like https://github.com/tidymodels/rules/issues/49, https://github.com/tidymodels/bonsai/issues/30, and others as they come up in the future.

tests/testthat/test-checks.R
```diff
@@ -179,6 +179,54 @@ test_that("workflow objects", {
...
 })
 
+test_that("workflow objects (will not tune, tidymodels/tune#548)", {
+  skip_if_not_installed("glmnet")
+
+  bare_rec <-
+    recipes::recipe(ridership ~ ., data = head(Chicago))
+
+  # don't warn when supplied tune args make sense given engine / steps
+  expect_silent(
```
DavisVaughan, line 189: Maybe use `expect_warning(code, NA)` here until we have `expect_no_warning()` in testthat

DavisVaughan:
And this is probably an error to look into too, possibly for another PR

``` r
library(tidymodels)

pset <- dials::parameters(dials::penalty())

tune::tune_grid(
  linear_reg(penalty = tune()),
  mpg ~ .,
  param_info = pset,
  bootstraps(mtcars)
)
#> Error in `compute_grid_info()` at tune/R/grid_code_paths.R:147:2:
#> ! Internal error: `workflow` should have some tunable parameters if `grid` is not `NULL`.
```

R/checks.R
```diff
@@ -33,12 +33,16 @@ check_grid <- function(grid, workflow, pset = NULL) {
...
-      "No tuning parameters have been detected, ",
-      "performance will be evaluated using the resamples with no tuning. ",
-      "Did you want to [tune()] parameters?"
+    msg <- c(
+      "!" = "No tuning parameters have been detected;
+             performance will be evaluated using the resamples with no tuning."
     )
-    rlang::warn(msg)
+
+    if (nrow(pset) == nrow(tune_args(workflow))) {
```
DavisVaughan, line 41:
```suggestion
    if (nrow(tune_args(workflow)) == 0L) {
```

Maybe slightly clearer since you are already in the `nrow(pset) == 0L` branch

R/checks.R
```diff
@@ -286,6 +290,24 @@ check_workflow <- function(x, pset = NULL, check_dials = FALSE) {
     }
   }
 
+  tune_tbl <- tune_args(x)
+
+  if (nrow(tune_tbl) > nrow(pset)) {
+    not_tunable <- tune_tbl$id[!tune_tbl$id %in% pset$id]
+    not_tunable_tbl <- tune_tbl[tune_tbl$id %in% not_tunable, ]
```
DavisVaughan, line 297:
```suggestion
    not_tunable_tbl <- tune_tbl[!tune_tbl$id %in% pset$id, ]
    not_tunable <- not_tunable_tbl$id
```

Maybe?

R/checks.R
```diff
@@ -286,6 +290,24 @@ check_workflow <- function(x, pset = NULL, check_dials = FALSE) {
...
+    msg <-
+      c("!" = "{cli::qty(not_tunable)}The parameter{?s} {.var {not_tunable}}
+               {?was/were} marked with `tune()`, though {?is/are} not supported for tuning.")
+    if (all(not_tunable_tbl$source == "model_spec")) {
+      msg <-
+        c(msg,
+          "i" = "{cli::qty(not_tunable)}Have you supplied an engine that
+                 supports tuning {?this/these} parameter{?s}?")
```
DavisVaughan, line 305: Can we avoid phrasing this as a question? It seems somewhat ambiguous because it can be interpreted either way

R/checks.R
```diff
@@ -286,6 +290,24 @@ check_workflow <- function(x, pset = NULL, check_dials = FALSE) {
...
 
+  tune_tbl <- tune_args(x)
+
+  if (nrow(tune_tbl) > nrow(pset)) {
+    not_tunable <- tune_tbl$id[!tune_tbl$id %in% pset$id]
+    not_tunable_tbl <- tune_tbl[tune_tbl$id %in% not_tunable, ]
+    msg <-
+      c("!" = "{cli::qty(not_tunable)}The parameter{?s} {.var {not_tunable}}
+               {?was/were} marked with `tune()`, though {?is/are} not supported for tuning.")
+    if (all(not_tunable_tbl$source == "model_spec")) {
```
DavisVaughan, line 301: Maybe change this to `any()` rather than `all()`? i.e. if at least 1 parameter isn't tunable but came from the model specification, then it is probably worth talking about it

You should filter `not_tunable` down to `not_tunable[not_tunable_tbl$source == "model_spec"]` for this if you change it to `any()`

simonpcouch:
Making a few changes here at the moment, but a proposal: I think we may want to elevate _the first warning_ to an error.

``` r
tune::tune_grid(
    parsnip::linear_reg(penalty = tune(), mixture = tune()),
    mpg ~ .,
    rsample::bootstraps(mtcars, 2)
)
#> Warning: ! The parameters `penalty` and `mixture` were marked with `tune()`, though will
#>   not be tuned.
#> ℹ This usually means that the current modeling engine `lm` does not support
#>   tuning `penalty` and `mixture`.
#> Warning: ! No tuning parameters have been detected; performance will be evaluated using
#>   the resamples with no tuning.
#> # Tuning results
#> # Bootstrap sampling 
#> # A tibble: 2 × 4
#>   splits          id         .metrics         .notes          
#>   <list>          <chr>      <list>           <list>          
#> 1 <split [32/13]> Bootstrap1 <tibble [2 × 4]> <tibble [0 × 3]>
#> 2 <split [32/10]> Bootstrap2 <tibble [2 × 4]> <tibble [0 × 3]>
```

<sup>Created on 2022-09-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

If a user supplies parameters for tuning, and sees the first warning, they will almost surely rerun this code. Fitting to resamples seems like a costly fallback, especially since the warning is only thrown when the results are returned, i.e. after the user has waiting for fitting to resamples.

If a user doesn't supply parameters for tuning but calls `tune::tune_grid`, I think it's fine if we continue to warn and fallback to fitting resamples. I believe was the original intention of scope for the second warning and I see its usefulness.

This would be breaking for code that relies on that first kind of fallback, that we unintentionally supported, which surely exists in some places, though I imagine not in package code.

R/checks.R
simonpcouch: Addressed in [84d08ed](https://github.com/tidymodels/tune/pull/549/commits/84d08ed54016169ebfd30597de11023801771849)!

R/checks.R
simonpcouch: Addressed in [84d08ed](https://github.com/tidymodels/tune/pull/549/commits/84d08ed54016169ebfd30597de11023801771849)!

tests/testthat/test-checks.R
simonpcouch: Addressed in [14faaa2](https://github.com/tidymodels/tune/pull/549/commits/14faaa2edc1897fc33ad1a3397ed315a9da519f6)!

R/checks.R
```diff
@@ -286,7 +286,25 @@ check_workflow <- function(x, pset = NULL, check_dials = FALSE) {
...
+    msg <-
+      c("!" = "{cli::qty(not_tunable)}The parameter{?s} {.var {not_tunable}}
+               {?was/were} marked with `tune()`, though will not be tuned.",
+        "i" = "This usually means that the current modeling engine
+               {.var {extract_spec_parsnip(x)$engine}}
+               does not support tuning {.var {not_tunable}}."
+      )
+
+
+    cli::cli_abort(msg, call = rlang::caller_env(2), class = "not_tunable_error")
```
simonpcouch, line 305: `rlang::caller_env(2)` always assumes that `check_workflow` will be called as `tune_*(tune_*_workflow(check_workflow()))`, which is currently the case, but may not be in the future. If it's called top-level or nested within only one call, we just get `"Error:"` as the header.

R/checks.R
```diff
@@ -286,7 +286,25 @@ check_workflow <- function(x, pset = NULL, check_dials = FALSE) {
     }
   }
 
-  mod <- extract_spec_parsnip(x)
+  mod <- hardhat::extract_spec_parsnip(x)
+
+  to_be_tuned <- hardhat::extract_parameter_set_dials(mod)
```
simonpcouch, line 291: Note that this is a separate parameter set from the one used earlier in the function. Is not sensitive to the user's `param_info` input and called only with respect to the model spec, rather than the workflow.

simonpcouch (commented):
This PR should be ready to go. :)

PR #554, opened 2022-09-28
add example in `tune_bayes()` docs

simonpcouch:
Run conditionally with `identical(Sys.getenv("NOT_CRAN"), "true")` so as not to increase CRAN check time.🐬

### DIFF
--- a/R/tune_bayes.R
+++ b/R/tune_bayes.R
@@ -127,6 +127,46 @@
 #'
 #' @inheritSection tune_grid Extracting Information
 #'
+#' @examplesIf (identical(Sys.getenv("NOT_CRAN"), "true") && rlang::is_installed("kernlab"))
+#' library(recipes)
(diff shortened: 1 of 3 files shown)


DavisVaughan:
We saw that ps has `ps:::is_cran_check()` for this kind of thing

```r
> ps:::is_cran_check
function () 
{
    if (identical(Sys.getenv("NOT_CRAN"), "true")) {
        FALSE
    }
    else {
        Sys.getenv("_R_CHECK_PACKAGE_NAME_", "") != ""
    }
}
```

topepo:
Should we add a version of `ps:::is_cran_check()` that also has an option to check for installed packages (e.g. ones from Suggests)?

simonpcouch:
That sounds like a good idea! I'll add that function and implement it for this example for this PR, and then implement it elsewhere in a separate one.

PR #562, opened 2022-10-13
use getter to extract condition message

simonpcouch:
Instead of subsetting conditions with `$`, this PR proposes using `conditionMessage` to retrieve the message from errors/warnings. The `$message` slot can be a vector of arbitrary length, while `conditionMessage` will collapse messages to length 1. 

[EDIT: no new issues in finetune as a result of this PR.]

### DIFF
--- a/R/grid_code_paths.R
+++ b/R/grid_code_paths.R
@@ -538,7 +538,7 @@ super_safely <- function(fn) {
 
   # Construct a try()-error to be compatible with `log_problems()`
   handle_error <- function(e) {
-    e <- structure(e$message, class = "try-error", condition = e)
+    e <- structure(conditionMessage(e), class = "try-error", condition = e)
     list(result = NULL, error = e, warnings = warnings)
   }
 
--- a/R/logging.R
+++ b/R/logging.R
@@ -120,7 +120,7 @@ log_problems <- function(notes, control, split, loc, res, bad_only = FALSE) {
 
   wrn <- res$signals
   if (length(wrn) > 0) {
-    wrn_msg <- purrr::map_chr(wrn, ~ .x$message)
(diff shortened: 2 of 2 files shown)


R/logging.R
```diff
@@ -120,7 +120,7 @@ log_problems <- function(notes, control, split, loc, res, bad_only = FALSE) {
 
   wrn <- res$signals
   if (length(wrn) > 0) {
-    wrn_msg <- purrr::map_chr(wrn, ~ .x$message)
+    wrn_msg <- purrr::map_chr(wrn, ~conditionMessage(.x))
```
simonpcouch, line 123: This was the underlying issue with https://github.com/tidymodels/finetune/pull/50. This function is called in the code path for handling logging cases with warnings but no errors. The condition message, in the case of the lifecycle message raised there (which is why this only occurred in `test()`—lifecycle messages aren't raised interactively when not called by the user), had length greater than 2. `map_chr()` thus errored out, which caused `log_problems()` to error out, which is not inside of the `.catch_and_log` code path that effectively handles user errors.

R/grid_code_paths.R
```diff
@@ -538,7 +538,7 @@ super_safely <- function(fn) {
 
   # Construct a try()-error to be compatible with `log_problems()`
   handle_error <- function(e) {
-    e <- structure(e$message, class = "try-error", condition = e)
+    e <- structure(conditionMessage(e), class = "try-error", condition = e)
```
simonpcouch, line 541: This was unrelated to the finetune issue, but if the logic for the anologous fix there is sound, I figured we'd apply to the error place where this comes up!

topepo (approved):

PR #572, opened 2022-10-18
accommodate metrics with a target value of zero

### LINKED ISSUE #243: metrics with a target value of zero
`show_best()` and others have code that looks like: 

```r
  if (maximize) {
    summary_res <- summary_res %>% dplyr::arrange(dplyr::desc(mean))
  } else {
    summary_res <- summary_res %>% dplyr::arrange(mean)
  }
```

This should be adjusted for cases when there is a target value. 

Perhaps we can use `maximize = FALSE` and arrange by `(abs(mean - target))`. 

@DavisVaughan

simonpcouch:
Closes #243!

### DIFF
--- a/R/select_best.R
+++ b/R/select_best.R
@@ -197,14 +197,14 @@ select_by_pct_loss.tune_results <- function(x, ..., metric = NULL, limit = 2) {
 
   res <-
     res %>%
-    rowwise() %>%
+    dplyr::rowwise() %>%
     dplyr::mutate(
       .best = best_metric,
       # can calculate loss without knowledge of direction since
       # we know that losses must be larger than that for the best
       .loss = abs((abs(mean) - abs(.best)) / .best) * 100
     ) %>%
-    ungroup()
+    dplyr::ungroup()
 
   res <- try(dplyr::arrange(res, !!!dots), silent = TRUE)
   if (inherits(res, "try-error")) {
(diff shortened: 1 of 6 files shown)


tests/testthat/_snaps/select_best.md
```diff
@@ -16,15 +16,15 @@
     Code
       select_best(rcv_results, metric = "random")
     Condition
-      Error in `is_metric_maximize()`:
+      Error in `get_metric_direction()`:
```
simonpcouch, line 19: These messages ought to call out the user-called function rather than the checking function. I went ahead and updated snaps without making changes to the erroring code, as correcting this will require some more invasive changes to pass around calling environments that I'll implement in a separate PR.

R/select_best.R
```diff
@@ -277,6 +281,20 @@ select_by_one_std_err.tune_results <- function(x, ..., metric = NULL) {
         .bound = bound
       ) %>%
       dplyr::filter(mean <= .bound)
+  } else if (direction == "zero") {
+    best_index <- which.min(abs(res$mean))
+    best <- res$mean[best_index]
+    bound_lower <- -abs(best) - res$std_err[best_index]
+    bound_upper <- abs(best) + res$std_err[best_index]
```
simonpcouch, line 288: A second set of eyes on these bound calculations would be appreciated—I believe this is the approach we want to take, though I'm not sure if we need to update any docs to clarify the definition of this bound.

PR #586, opened 2022-12-05
fit_best()

topepo:
PR for generating a model fit from tuning results using the numerically best tuning parameter configurations. 

My use of cli for the messaging could be better, so please suggest improvements.


R/fit_best.R
```diff
@@ -0,0 +1,140 @@
...
+#'
+#' The data used for the fit are taken from the `splits` column. If the split
+#' column was from a validation split, the combined training and validation sets
+#' are used.
+#'
+#' In comparison to [last_fit()], that function requires a finalized model, fits
+#' the model on the data set by [rsample::initial_split()], and computes
+#' metrics from the test set.
+#' @examples
+
```
EmilHvitfeldt, line 31: remove empty line

R/fit_best.R
```diff
@@ -0,0 +1,140 @@
...
+fit_best.default <- function(x, ...) {
+  cls <- paste0("'", class(x), "'", collapse = ", ")
+  msg <-
+    paste("There is no `fit_best()` method for an object with class(es):", cls)
+  rlang::abort(msg)
+}
+
+#' @export
+#' @rdname fit_best
+fit_best.tune_results <- function(x, metric = NULL, parameters = NULL, verbose = FALSE, ...) {
```
EmilHvitfeldt, line 77:
```suggestion
fit_best.tune_results <- function(x, 
                                  metric = NULL, 
                                  parameters = NULL, 
                                  verbose = FALSE, 
                                  ...) {
```

Wrap the arguments in Function-indent style

R/fit_best.R
```diff
@@ -0,0 +1,140 @@
...
+  if (length(list(...))) {
+    cli::cli_abort(c("x" = "The `...` are not used by this function."))
+  }
+  wflow <- .get_tune_workflow(x)
+  if (is.null(wflow)) {
+    cli::cli_abort(c("x" = "The control option `save_workflow = TRUE` should be used when tuning."))
```
EmilHvitfeldt, line 83: This line is too long. One suggested change is the following. Another would be to use `paste()`

```suggestion
msg <- 
      "The control option `save_workflow = TRUE` should be used when tuning."
    cli::cli_abort(c("x" = msg))
```

R/fit_best.R
```diff
@@ -0,0 +1,140 @@
...
+
+  wflow <- finalize_workflow(wflow, parameters = parameters)
+
+  # ----------------------------------------------------------------------------
+
+  pset <- hardhat::extract_parameter_set_dials(wflow)
+  p <- nrow(pset)
+  if (p > 0) {
+    prms <- paste0("'", pset$id, "'", collapse = ", ")
+    msg <- paste("There {?is/are} {p} parameter{?s} still marked for tuning: ", prms)
```
EmilHvitfeldt, line 106: This line is also a little too long

R/fit_best.R
```diff
@@ -0,0 +1,140 @@
...
+    cli::cli_abort(
+      msg
+    )
+  }
+
+  # ----------------------------------------------------------------------------
+
+  dat <- x$splits[[1]]$data
+  if (verbose) {
+    cli::cli_inform(paste("Fitting using", nrow(dat), "data points..."))
```
EmilHvitfeldt, line 116: Are the `...` necessary? it is not immediately obvious to me (non-native speaker) what this is signaling

R/fit_best.R
```diff
@@ -0,0 +1,140 @@
...
+  }
+  res <- parsnip::fit(wflow, dat)
+  if (verbose) {
+    cat(cli::cli_inform("done\n\n"))
```
EmilHvitfeldt, line 120:
```suggestion
  cli::cli_inform("done")
```
Did you need `cat()` and `\n\n`?

EmilHvitfeldt (commented):
couple of notes on my first read-through

R/fit_best.R
simonpcouch: cli also allows for line-breaking in code without inserting line breaks into results, e.g.

```
cli::cli_abort("a very very very \\
                long message")
#> Error:
#> ! a very very very long message
```

simonpcouch:
Heads up: will go ahead and pull + make edits to resolve Emil's review comments and workshop cli calls!

R/fit_best.R
simonpcouch: I read the ellipses as indicating that the action messaged about is in progress rather than completed, but worth considering if you tripped up on this! Will leave unresolved for Max.

R/fit_best.R
```diff
@@ -0,0 +1,147 @@
...
+#'   wflow <- finalize_workflow(wflow, best_param)  # or just `finalize_model()`
+#'   wflow_fit <- fit(wflow, data_set)
+#' }
+#'
+#' The data used for the fit are taken from the `splits` column. If the split
+#' column was from a validation split, the combined training and validation sets
+#' are used.
+#'
+#' In comparison to [last_fit()], that function requires a finalized model, fits
+#' the model on the data set by [rsample::initial_split()], and computes
```
simonpcouch, line 32: I think the phrasing

> "fits the model on the data set by [rsample::initial_split()]" 

could use some clarification.

simonpcouch (approved):
Handing off! Really excited about this addition.

R/fit_best.R
topepo: Also, before I was using cli was just using cat. I would cat the first message, ending in the three dots and then write a 'done' on the same line if everything succeeded with the model fit.

Just so you know why those were originally there

R/fit_best.R
topepo: How about "the training set defined by" the initial split?

simonpcouch: That sounds great!

PR #587, opened 2022-12-07
workshop pluralization in `filter_parameters()`

simonpcouch:


R/filter_parameters.R
```diff
@@ -80,11 +80,11 @@ filter_by_join <- function(x, parameters = NULL, nm = "") {
   }
   extra_names <- setdiff(filter_names, param_names)
   if (length(extra_names) > 0) {
-    msg <- paste0(
-      "There are unneeded columns in `parameters` that were ignored: ",
-      paste0("'", extra_names, "'", collapse = ", ")
-    )
-    rlang::warn(msg)
+    cli_warn(c(
```
hfrick, line 83: I think you don't need the `c()` here

R/filter_parameters.R
```diff
@@ -80,11 +80,11 @@ filter_by_join <- function(x, parameters = NULL, nm = "") {
...
+      "{qty(extra_names)} The column{?s} {.var {extra_names}} passed in \\
+       {.arg parameters} {?is/are} unneeded and will be ignored."
```
hfrick, line 85: 🎉

R/filter_parameters.R
```diff
@@ -80,11 +80,11 @@ filter_by_join <- function(x, parameters = NULL, nm = "") {
...
+    ))
```
hfrick, line 86: it'd be nice to pass the call through from `filter_parameters()`

hfrick (commented):
Love to see it! I've left some detailed comments for your discretion since the PR was so small 😄

R/filter_parameters.R
simonpcouch: You're right! Deleted in [80cfe61](https://github.com/tidymodels/tune/pull/587/commits/80cfe61004a3d4ca6494444181f1554e60e86aed).

R/filter_parameters.R
simonpcouch: Agreed, though unfortunately `rlang::warn()` (via `cli_warn`) doesn't accept a `call` argument in the way that `rlang::abort()` (via `cli_abort()`) does. :/

hfrick: Oh, I see! Do you know if that's intentional?

simonpcouch: I'm not sure what the logic is there. I was wrong here, though: looks like there is indeed something happening the ellipses:

``` r
rlang::warn("boop")
#> Warning: boop
rlang::warn("boop", call = "bop")
#> Warning in "bop": boop
```

<sup>Created on 2022-12-09 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Will update in a moment!

simonpcouch: Oh, but it looks like the machinery in that argument doesn't work quite the same.

With `caller_env()` in this warnings `call`, we see:

```
Warning message:
In filter_parameters(svm_reg_results, parameters = tibble::tibble(`%^*#` = 1,  :
  The column `soup` passed in `parameters` is not needed and will be ignored.
```

Issue #588, opened 2022-12-08
catalog tuning issues interactively


simonpcouch:
This issue will link related PRs that implement the issue cataloger. Rather than printing warnings/errors as they come up, the cataloger collapses information on issues with fitting to resamples (a la existing `tune:::summarize_notes()` used in `print.tune_results()`) for less overwhelming output when things go wrong:

https://user-images.githubusercontent.com/35748691/208160475-dadb5b22-432d-4a30-a349-b3bd83c466ac.mp4

This feature requires a few small internal tweaks that, for ease of reviewing and documenting, I'll PR separately before the cataloger-specific changes!

PR #589, opened 2022-12-08
separate `get_operator()` into two functions

simonpcouch:
The issue cataloger (#588) only kicks in when parallelism will not be used (i.e. when `tune:::get_operator()` returns `foreach::%do%` rather than `foreach::%dopar%`). This PR makes no functional changes to `get_operator()`, but splits the logic that generates the logical value that `get_operator()` will use to conditionally pick the `do` operator into a helper called `allow_parallelism()`. We can then use the conditions generated by `allow_parallelism()` to decide whether to raise prompts as usual or use the issue cataloger.

### DIFF
--- a/R/parallel.R
+++ b/R/parallel.R
@@ -2,23 +2,31 @@
 # Helpers for parallel processing
 
 # object should be a workflow
-get_operator <- function(allow = TRUE, object) {
+allow_parallelism <- function(allow = TRUE, object = NULL) {
   is_par <- foreach::getDoParWorkers() > 1
-  pkgs <- required_pkgs(object)
-  blacklist <- c("keras", "rJava")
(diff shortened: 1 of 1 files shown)


EmilHvitfeldt (approved):
Looks good!

R/parallel.R
```diff
@@ -2,23 +2,31 @@
...
   }
 
-  cond <- allow && is_par
+  allow && is_par
+}
+
+get_operator <- function(allow = TRUE, object) {
+  cond <- allow_parallelism(allow, object)
+
   if (cond) {
```
EmilHvitfeldt, line 25: Consider the following:

```suggestion
  if (allow_parallelism(allow, object)) {
```

I feel like it reads pretty well

simonpcouch: I'm for it!

PR #590, opened 2022-12-08
allow existing exit handlers to persist in `tune_bayes()`

simonpcouch:
Related to #588. Currently, the `on.exit()` calls in `tune_bayes_workflow()` act on the function's execution environment. This means that any exit handlers on the function env that exist before the first call to `on.exit()` will be cleared then:

https://github.com/tidymodels/tune/blob/3f1b4a7ec162305b17175d5274fba887980ca629/R/tune_bayes.R#L281-L294

...and then won't be restored on the second call:

https://github.com/tidymodels/tune/blob/3f1b4a7ec162305b17175d5274fba887980ca629/R/tune_bayes.R#L422-L423

This PR wraps the portion of the code that relies on that exit handler in `evalq` so that existing exit handlers (to be introduced by the cli progress bar in a future PR) persist.


hfrick (approved):
👍

PR #593, opened 2022-12-16
log tuning issues interactively

### LINKED ISSUE #588: catalog tuning issues interactively

simonpcouch:
Closes #588. :)

Refines machinery for logging issues during tuning. Rather than printing out warnings and errors as they appear, this PR proposes only printing unique tuning issues and iteratively updating a progress bar to maintain counts of each unique issue. This feature is only enabled for tuning sequentially.

The "issue cataloger" makes use of a `tune_env` environment object that lives in tune's namespace and allows for flexibly passing around details about issues encountered during a tuning process. The values of `progress_*` objects inside of it are attached to exit handlers that look to the execution environment of the lowest-level function that defines a tuning process, e.g. `tune_grid_loop_impl()` for `tune_grid()` and `fit_resamples()` or `tune_bayes_workflow()` for `tune_bayes()`.

To be followed up with a finetune PR here in a moment to place `initialize_catalog()` in the right places, and then one to extratests with testing for finetune and agua cataloging.


simonpcouch:
The last couple passing CI runs have required re-running one or more GHA workflows to successfully output the expected cli progress bar matches. I’m not sure we can reliably test this feature on CI.🤔

R/logging.R
```diff
@@ -63,6 +63,157 @@ message_wrap <-
...
+  )
+
+# determines whether a supplied bar id references a tuning process that
+# uses the issue cataloger.
+uses_catalog <- function() {
+  isTRUE(tune_env$progress_active && !is_testing())
+}
+
+# determines whether a supplied bar id references an issue cataloger that
+# has not yet been terminated. this function is intended to guard calls to
```
DavisVaughan, line 79: I think you need to update these comments? i.e. they reference a "supplied bar id" but there aren't arguments to this anymore

R/logging.R
```diff
@@ -63,6 +63,157 @@ message_wrap <-
...
+catalog_is_active <- function() {
+  tune_env$progress_active
+}
+
+# initializes machinery for the tune catalog inside of an environment.
+# the `env` should be an execution environment that persists throughout the
+# tuning process for a given tuning approach and exits once tuning is completed.
+#' @rdname tune-internal-functions
+#' @export
+initialize_catalog <- function(env = rlang::caller_env(), control) {
```
DavisVaughan, line 92: I'd reverse the order of the arguments since one is optional and not normally specified

R/logging.R
```diff
@@ -63,6 +63,157 @@ message_wrap <-
...
+# printing out each new tuning issue individually.
+tune_catalog <- function(issues) {
+  catalog <- rlang::env_get(env = tune_env, nm = "progress_catalog")
+
+  res <- dplyr::count(issues, type, note) %>% mutate(id = NA_integer_)
+  res <- dplyr::bind_rows(res, catalog)
+  res <- dplyr::group_by(res, type, note)
+  # dplyr::first will gain an `na_rm` argument in 1.1.0
+  res <- dplyr::summarize(res, n = sum(n), id = dplyr::first(id[!is.na(id)]))
+  res <- dplyr::ungroup(res)
```
DavisVaughan, line 175: Do you want to leave it grouped by `type` and `note`?

R/logging.R
```diff
@@ -63,6 +63,157 @@ message_wrap <-
...
+
+  for (issue in seq_along(res[[1]])) {
```
DavisVaughan, line 177:
```suggestion
  for (issue in seq_len(nrow(res))) {
```

DavisVaughan (approved):
I think it would be nice if Max also reviewed this and tried it out locally on a number of examples

If the tests are flaky, we definitely don't want to merge them. They will just cause us a lot of pain and wasted time. It is possible we just can't test this well due to cli being a bit "random" about when it emits messages. We might want to just pull the tests into a separate script that we keep around to run interactively to manually "test" that things are working as expected if we come back and touch this code again

R/logging.R
simonpcouch: Ah, thank you!

R/logging.R
simonpcouch: You got it.

R/logging.R
simonpcouch: Thanks!

R/logging.R
simonpcouch: I don't believe so. I don't think we rely on grouping / re`group_by()` later on, so I figure removing groups is the more cautious thing to do.

simonpcouch:
That sounds like a good move re: testing. In [cf1d5dd](https://github.com/tidymodels/tune/pull/593/commits/cf1d5dd62a82beaabed4c9a6460863ff0bf9076c), I transitioned those testthat tests to a script in `inst/` and added an entry to `release_bullets()` to take a look at them.

simonpcouch:
@topepo, no need to look at this sooner than later, but we thought you might want to push this around before merging.

Tests are failing pending #595, and those will fail until GHA can successfully install the new purrr release.

R/logging.R
DavisVaughan: Oh I was totally just reading this wrong. Completely ignore this, I think what you have is right

simonpcouch: Ah, okay! Sounds good.

R/logging.R
```diff
@@ -63,6 +63,156 @@ message_wrap <-
...
+    return("")
+  }
+
+  res <- dplyr::arrange(catalog, id)
+  res <- dplyr::mutate(res, color = dplyr::if_else(type == "warning", list(cli::col_yellow), list(cli::col_red)))
+  res <- dplyr::rowwise(res)
+  res <- dplyr::mutate(res, msg = glue::glue("{color(cli::style_bold(id))}: x{n}"))
+  res <- dplyr::ungroup(res)
+  res <- dplyr::pull(res, msg)
+  res <- glue::glue_collapse(res, sep = "   ")
```
topepo, line 137: Maybe make the number of spaced an argument? If we eventually nest this inside of another function we might want different amounts of whitespace to differentiate nested catalogs?

simonpcouch: You got it! [2e94742](https://github.com/tidymodels/tune/pull/593/commits/2e94742c45a469776182246da0dcbd3ca9e92db6)

PR #594, opened 2022-12-16
transition messaging functions

simonpcouch:
Related to #588. `cli::cli_bullets()` is aware of the progress bar and will clear it before printing, while `rlang::inform()` will print on top of it and leave remaining characters visible:

https://user-images.githubusercontent.com/35748691/208186327-0fe06dd5-d2d1-4dec-91b5-190d8648f20e.mov

### DIFF
--- a/tests/testthat/_snaps/bayes.md
+++ b/tests/testthat/_snaps/bayes.md
@@ -451,7 +451,8 @@
       tune:::check_bayes_initial_size(5, 3, FALSE)
     Message
       ! There are 5 tuning parameters and 3 grid points were requested.
-      * There are more tuning parameters than there are initial points. This is likely to cause numerical issues in the first few search iterations.
+      * There are more tuning parameters than there are initial points. This is
+        likely to cause numerical issues in the first few search iterations.
 
 ---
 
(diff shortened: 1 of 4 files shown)


tests/testthat/_snaps/bayes.md
```diff
@@ -451,15 +451,17 @@
...
+        likely to cause numerical issues in the first few search iterations.
 
 ---
 
     Code
       tune:::check_bayes_initial_size(5, 3, TRUE)
     Message
       ! There are 5 tuning parameters and 3 grid points were requested.
-      * There are more tuning parameters than there are initial points. This is likely to cause numerical issues in the first few search iterations.
+      * There are more tuning parameters than there are initial points. This is
```
EmilHvitfeldt, line 463: Can you explain to me why this line being split in 2?

EmilHvitfeldt (approved):
One small question. Otherwise looks good!

tests/testthat/_snaps/bayes.md
simonpcouch: A bit buried in the docs, but from the [cli pkgdown site](https://cli.r-lib.org/reference/inline-markup.html?q=break#wrapping):

> Most cli containers wrap the text to width the container's width, while observing margins requested by the theme.

So, `rlang::inform` output is wrapped only if supported by the console, `cli::cli_bullet` output is wrapped in more contexts.🎁

PR #595, opened 2022-12-21
update snaps with purrr 1.0.0

simonpcouch:


simonpcouch:
Currently seeing:

```
Error: Error: package or namespace load failed for 'tune' in library.dynam(lib, package, package.lib):
Error: Error: R CMD check found ERRORs
 DLL 'purrr' not found: maybe not installed for this architecture?
Error: Error: loading failed
```

so will need to wait a couple days.🏄

PR #596, opened 2023-01-20
iterate labels with letters rather than numbers in cataloger

simonpcouch:
Following the recommendation in https://github.com/tidymodels/finetune/pull/60, this PR iterates labels with letters rather than numbers in the cataloger.

What was:
```
#> → 1 | warning: ope! yikes.
#> → 2 | error: AHHhH
#> There were issues with some computations   1: x10   2: x10
```

...is now:

```
#> → A | warning: ope! yikes.
#> → B | error: AHHhH
#> There were issues with some computations   A: x10   B: x10
```

I agree that this is more readable.🙂

Rather than putting together machinery to handle the newly needed edge case of running out of labels, I opted to pad the vector of labels with a sequence of 1000 numbers:

```
lbls <- c(LETTERS, letters, 1:1e3)
```

### DIFF
--- a/R/logging.R
+++ b/R/logging.R
@@ -69,6 +69,8 @@ tune_env <-
     data = list(progress_env = NULL, progress_active = FALSE, progress_catalog = NULL)
   )
 
+lbls <- c(LETTERS, letters, 1:1e3)
+
 # determines whether a currently running tuning process uses the cataloger.
 uses_catalog <- function() {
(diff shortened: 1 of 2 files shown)


simonpcouch:
Snap failures look unrelated to this PR: https://github.com/tidymodels/tune/issues/597

topepo (approved):

PR #619, opened 2023-02-22
incorporate an evaluation time argument for censored regression

topepo:
First of a few censored regression PRs. 

I won't merge this in until after the next tune release since it will take some time (and a yardstick release) to get right.


R/tune_bayes.R
```diff
@@ -192,7 +192,8 @@ tune_bayes.model_spec <- function(object,
                                   metrics = NULL,
                                   objective = exp_improve(),
                                   initial = 5,
-                                  control = control_bayes()) {
+                                  control = control_bayes(),
+                                  eval_times = NULL) {
```
topepo, line 196: How do we feel about `eval_times` as an argument?

EmilHvitfeldt: I like it, not too overly long

DESCRIPTION
```diff
@@ -56,5 +57,8 @@ Config/testthat/edition: 3
 Encoding: UTF-8
 Language: en-US
 LazyData: true
+remotes: 
+    tidymodels/yardstick#347,
+    r-lib/rlang
```
topepo, line 62: This is to fix a bad link in a snapshot. Lionel thinks it will be on CRAN within 2 weeks.

R/checks.R
```diff
@@ -346,6 +346,9 @@ check_metrics <- function(x, object) {
            classification = {
              x <- yardstick::metric_set(roc_auc, accuracy)
            },
+           'censored regression' = {
+             x <- yardstick::metric_set(yardstick::brier_survival)
```
topepo, line 350: We may update this later to include another metric.

R/grid_helpers.R
```diff
@@ -102,6 +95,37 @@ predict_model <- function(split, workflow, grid, metrics, submodels = NULL) {
   tibble::as_tibble(res)
 }
 
+predict_wrapper <- function(model, new_data, type, times, subgrid = NULL) {
```
topepo, line 98: This funciton cleans up the code where it is used (a little) and would allow us to add more hooks for future special cases.

R/grid_performance.R
```diff
@@ -46,7 +48,15 @@ metrics_info <- function(x) {
 #' @keywords internal
 #' @name tune-internal-functions
 #' @export
-.estimate_metrics <- function(dat, metric, param_names, outcome_name, event_level) {
+.estimate_metrics <- function(dat, metric, param_names, outcome_name, event_level, stat_time = NULL) {
+  # The call stack is:
```
topepo, line 52: Just some notes that I'll remove later

R/checks.R
```diff
@@ -357,27 +360,36 @@ check_metrics <- function(x, object) {
...
-      "but `metrics` is a metric set for regression metrics."
+      "but `metrics` is a metric set for a different model mode."
     )
     rlang::abort(msg)
   }
 
+  if (mode == "censored regression" && !is_surv_metric_set) {
+    msg <- paste0(
+      "The parsnip model has `mode = 'censored regression'`, ",
+      "but `metrics` is a metric set for other model modes."
```
simonpcouch, line 389:
```suggestion
      "but `metrics` is a metric set for a different model mode."
```

Just for consistency with the others.

R/checks.R
simonpcouch, line 350: Let's import this in the same way as those above

R/checks.R
```diff
@@ -598,3 +612,26 @@ check_no_tuning <- function(x) {
...
+
+check_eval_times <- function(.times, metrics) {
+  metric_types <- tibble::as_tibble(metrics)$class
+  has_dyn <- any(metric_types == "dynamic_survival_metric")
+  if (!is.null(.times) & !has_dyn) {
+    rlang::abort("evaluation times are only used for dynamic survival metrics.")
+  }
+  if (is.null(.times) & has_dyn) {
+    rlang::abort(
+      "1+ metric require the specification of time points in the `eval_times` argument."
```
simonpcouch, line 624:
```suggestion
      "One or more metric requires the specification of time points in the `eval_times` argument."
```

Same comment on setting `call`.

R/checks.R
```diff
@@ -598,3 +612,26 @@ check_no_tuning <- function(x) {
...
+    )
+  }
+
+  # will still propagate nulls
+  .times <- .times[!is.na(.times)]
+  .times <- unique(.times)
+  .times <- sort(.times)
+  .times <- .times[.times >= 0]
+  if (identical(.times, numeric(0))) {
+    rlang::abort("There were no usable evaluation times.")
```
simonpcouch, line 634: Same comment on `call`.

R/checks.R
```diff
@@ -598,3 +612,26 @@ check_no_tuning <- function(x) {
   rlang::abort(msg, call = NULL)
 }
 
+
+check_eval_times <- function(.times, metrics) {
+  metric_types <- tibble::as_tibble(metrics)$class
+  has_dyn <- any(metric_types == "dynamic_survival_metric")
+  if (!is.null(.times) & !has_dyn) {
+    rlang::abort("evaluation times are only used for dynamic survival metrics.")
```
hfrick:
```suggestion
    rlang::abort("Evaluation times are only used for dynamic survival metrics.")
```

R/checks.R
hfrick:
```suggestion
      "At least one metric requires the specification of evaluation time points in the `eval_times` argument."
```

R/tune_bayes.R
simonpcouch: On board!🙂

simonpcouch (commented):
Shaping up well!🏺 Will take care of a few of these changes myself and holler when I'm wrapped up.

R/checks.R
```diff
@@ -598,3 +612,30 @@ check_no_tuning <- function(x) {
...
+    )
+  }
+
+  # will still propagate nulls
+  .times <- .times[!is.na(.times)]
+  .times <- unique(.times)
+  .times <- sort(.times)
+  .times <- .times[.times >= 0]
+  if (identical(.times, numeric(0))) {
+    rlang::abort("There were no usable evaluation times.", call = NULL)
```
simonpcouch, line 638: Could this message provide some guidance (or link to docs) on what makes a usable evaluation time?

hfrick (commented):
Regarding the argument name: I generally like it but am wondering if `eval_time` would be better than `eval_times`, i.e. singular instead of plural. I am under the impression that we default to singular in such cases but am not sure about that and couldn't find a note in the design guide or style guide. It feels like something we may have a convention on though?

Regardless of which version we pick, it would be great to go over the various occasions this comes up in spirit under a different name. I see `eval_times`, `.times`, `times`, and `stat_time` and I think they all refer to the evaluation time.

R/resample.R
```diff
@@ -13,6 +13,9 @@
 #' @param control A [control_resamples()] object used to fine tune the resampling
 #'   process.
 #'
+#' @param eval_times A numeric vector of time points where dynamic event time
```
hfrick, line 16: this could be inherited from `tune_grid()`

R/checks.R
topepo: `eval_time` it is. I plan on going back and "harmonizing" the object/argument names. We might want to soft deprecate `predict(object, type = "survival", time)` to have a modifier.

R/resample.R
topepo: I noticed that and plan to do a small PR that maximizes inheritance across the `tune_*()` functions and `fit_resamples()`.

R/checks.R
hfrick: keeping track of that here: https://github.com/tidymodels/parsnip/issues/892

R/checks.R
hfrick: This looks like we might want to also avoid `Inf`

R/resample.R
topepo: See #622

topepo:
Death by lava on oldrel-2

```
  ℹ Building lava 1.7.2
  ℹ Building parsnip 1.0.4
  ✔ Built parsnip 1.0.4 (497ms)
  ℹ Building sessioninfo 1.2.2
  ✖ Failed to build lava 1.7.2
  Error: 
  ! error in pak subprocess
  Caused by error in `stop_task_build(state, worker)`:
  ! Failed to build source package 'lava'
  Full installation output:
  * installing *source* package ‘lava’ ...
  ** package ‘lava’ successfully unpacked and MD5 sums checked
  staged installation is only possible with locking
  ** using non-staged installation
  ** R
  Error in parse(outFile) : 
    /tmp/RtmpD2V0r5/R.INSTALL15a72bfd5a4f/lava/R/contr.R:47:53: unexpected input
  46: pairwise.diff <- function(n) {
  47:   pdiff <- function(n) lava::contr(lapply(seq(n-1), \
                                                          ^
  ERROR: unable to collate and parse R files for package ‘lava’
  * removing ‘/tmp/RtmpDaUYrW/pkg-lib134f324959ac/lava’
 ```

I'll restart later and it'll probably work.

**Comment by EmilHvitfeldt** [comment]
This PR can be moved to use dev version of yardstick instead of branch

**Comment by simonpcouch** [comment]
Working on getting these checks passing--will message when I'm handing off again!

**Comment by simonpcouch** [comment]
Okay, putting this away for now. `collect_metrics()` and `collect_predictions(summarize = FALSE)` seem to work correctly, but `collect_predictions(summarize = TRUE)` sees a `C stack usage` error. This comes up in `class_summarize()` rather than `collector()`, though:

https://github.com/tidymodels/tune/blob/7d8ce65a9630fa152f985c902522e0bce715c35c/R/collect.R#L320-L326

Not sure that this was yet a solved issue in previous commits, so leaving that alone.

File: R/grid_performance.R [hunk 12]
```diff
@@ -125,3 +138,69 @@ estimate_class_prob <- function(dat, metric, param_names, outcome_name,
...
+  # Other (non-submodel) parameters will be outside of 'x'.
+  # If this happens, pull the submodel parameters out of .pred and put them at
+  # the out level ('x') with the rest (if any)
+  outer_nms <- names(x)
+  inner_nms <- names(x$.pred[[1]])
+  has_inner_params <- any(params %in% inner_nms)
+  has_pred <- any(outer_nms == ".pred")
+  if (is.null(params) || !has_pred || !has_inner_params) {
+    return(x)
+  }
```
hfrick, line 168:
```suggestion
   if (is.null(params)) {
    return(x)
   }
   
  # When multi_predict() is used, .pred will have the tuning parameter values.
  # Other (non-submodel) parameters will be outside of 'x'.
  # If this happens, pull the submodel parameters out of .pred and put them at
  # the out level ('x') with the rest (if any)
  
  outer_nms <- names(x)
  has_pred <- any(outer_nms == ".pred")
   if (!has_pred) {
    return(x)
  }
  
  inner_nms <- names(x$.pred[[1]])
  has_inner_params <- any(params %in% inner_nms)
  if (!has_inner_params) {
    return(x)
  }

```

hfrick: @topepo this is where the warning `Unknown or uninitialised column: `.pred`.` came from. With this suggestion, the early returns are as early as they can be, hence not looking for `names(x$.pred[[1]])` when there is no `.pred` column.

PR #634, opened 2023-03-12
remove `tidyr_new_interface()` check

simonpcouch:
Cutting out complexity is nice! But also, turns out `packageVersion()` is a slow-poke.🦥

With main dev:

``` r
library(tidymodels)

bench::mark(
  fit = fit_resamples(linear_reg(), mpg ~ ., bootstraps(mtcars))
)
#> # A tibble: 1 × 6
#>   expression        min    median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>   <bch:tm>  <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 fit             1.11s     1.11s     0.907    30.5MB     7.25
```

With this PR:

``` r
#> # A tibble: 1 × 6
#> 1 fit           1.03s    1.03s     0.936    29.4MB     8.42
```

<sup>Created on 2023-03-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

The minimum version for tidyr in DESCRIPTION has been bumped since these changes were added anyway, so this helper would always return `TRUE`:

https://github.com/tidymodels/tune/blob/e61a83bf4180d38cd85234e5ab188a4579ca2fbd/DESCRIPTION#L36

### DIFF
--- a/R/0_imports.R
+++ b/R/0_imports.R
@@ -50,10 +50,6 @@ utils::globalVariables(
 
 # ------------------------------------------------------------------------------
 
-tidyr_new_interface <- function() {
-  utils::packageVersion("tidyr") > "0.8.99"
-}
-
 release_bullets <- function() {
   c(
     "Update dependencies with `devtools::install_dev_deps()` and update the test objects via `R CMD BATCH --vanilla inst/test_objects.R`.",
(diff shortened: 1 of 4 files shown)


hfrick (approved):
Nice! 🎈

PR #640, opened 2023-03-15
write out unneeded `nest()` call

simonpcouch:
We call `tidyr::nest()` on grid information for each resample (or, depending on `parallel_over`, each resample vs model vs preprocessor combination). Taken together, this nesting takes up a good amount of time, especially so for fits to many resamples and quickly-running model fits. 

In this case, writing that `nest()` call out saves us something like 8% of evaluation time. With `main` dev:

``` r
library(tidymodels)

bench::mark(
  res = fit_resamples(linear_reg(), mpg ~ ., bootstraps(mtcars, 100))
)
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
#> # A tibble: 1 × 6
#> 1 res           3.14s    3.14s     0.319    46.6MB     8.60
```

With this PR:

``` r
#> # A tibble: 1 × 6
#> 1 res           2.89s    2.89s     0.346    46.4MB     8.66
```

<sup>Created on 2023-03-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/grid_code_paths.R
+++ b/R/grid_code_paths.R
@@ -292,28 +292,12 @@ tune_grid_loop_iter <- function(split,
   model_param_names <- model_params$id
   preprocessor_param_names <- preprocessor_params$id
 
-  # Model related grid-info columns
-  cols <- rlang::expr(
-    c(
-      .iter_model,
-      .iter_config,
-      .msg_model,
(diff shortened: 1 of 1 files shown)


EmilHvitfeldt (approved):
Looks good!

PR #642, opened 2023-03-17
pre-compute `metrics_info(metrics)`

simonpcouch:
The `metrics_info` helper is a small helper to reformat `as_tibble.metric_set` for easier filtering/selecting. It's reasonably fast, but was called in a couple places inside the `tune_grid_loop_iter` _interior_ loops, so ends up being called thousands of times even for small grid searches. We can thus pre-compute that output and pass it as an argument to `tune_grid_loop_iter`. Note that this introduces merge conflicts with #641—glad to take care of that once we get there.

With `main` dev:

``` r
library(tidymodels)

set.seed(1)

bench::mark(
  total = fit_resamples(linear_reg(), mpg ~ ., bootstraps(mtcars, 100)),
  iterations = 5
)
#> # A tibble: 1 × 6
#> 1 total         2.71s    2.73s     0.363    46.3MB     6.96
```

...with a 4% decrease in evaluation time with this PR:

``` r
#> # A tibble: 1 × 6
#> 1 total          2.6s    2.62s     0.377    45.3MB     7.23
```

That effect is more substantial when tuning hyperparameters. :)

<sup>Created on 2023-03-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/grid_code_paths.R
+++ b/R/grid_code_paths.R
@@ -264,7 +264,7 @@ tune_grid_loop_iter <- function(split,
                                 metrics,
                                 control,
                                 seed,
-                                metrics_info = metrics_info(metrics)),
+                                metrics_info = metrics_info(metrics),
                                 params) {
   load_pkgs(workflow)
   .load_namespace(control$pkgs)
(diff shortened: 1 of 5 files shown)


EmilHvitfeldt (approved):
Amazing! Looking good

PR #655, opened 2023-03-31
transition from deprecated `dplyr::*_at()` variants

simonpcouch:
🥑

### DIFF
--- a/R/collect.R
+++ b/R/collect.R
@@ -271,6 +271,8 @@ prob_summarize <- function(x, p) {
   x
 }
 
+# define a `mean()` wrapper to avoid slowdown in
+# evaluation of anonymous function
 mean_na_rm <- function(x) {
   mean(x, na.rm = TRUE)
 }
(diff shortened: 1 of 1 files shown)


simonpcouch:
(Shamefully) tacking on [c642b5c](https://github.com/tidymodels/tune/pull/655/commits/c642b5cf1531b5d606837666753c40b3f4b8ba07) to this PR, as a separate PR would touch the same lines but would be a quick one. Use of an anonymous function drops into `dplyr:::as_inlined_function()`, where `quo()` machinery results in a pretty substantial slowdown.

R/collect.R
```diff
@@ -269,6 +271,10 @@ prob_summarize <- function(x, p) {
   x
 }
 
+mean_na_rm <- function(x) {
```
hfrick, line 274: Do you think it's worth it leaving a comment here that this is to avoid using an anonymous function for speed reasons?

hfrick (approved):
Good collection of commits 😜

R/collect.R
simonpcouch: Yes, good point.

simonpcouch: [658a1dc](https://github.com/tidymodels/tune/pull/655/commits/658a1dc5679792419f73c86460de2ad8be44e3ae)!

PR #671, opened 2023-04-14
reword relationship of `fit_best()` to `last_fit()`

hfrick:
I've found the wording mildly confusing so here is a different suggestion, happy to wordsmith it further if needed!


simonpcouch (approved):
On board!🍄

PR #672, opened 2023-04-14
error informatively when `predict()` is supplied tuning results

simonpcouch:
A little pit-of-success deepening with an eye for the [blog post](https://lsinks.github.io/posts/2023-04-10-tidymodels/tidymodels_tutorial.html) mentioning that the analogues for which prediction-related functions apply to which objects can be confusing.

The snaps look a bit different than how this feels interactively:

https://user-images.githubusercontent.com/35748691/232146348-60bd68ce-e255-4104-9e54-7abd4a96c927.mov

.

Another option could be to `deparse(substitute())` the input and then actually supply runnable code, like:

``` r
predict(ames_grid_search)
#> ! `predict()` is not well-defined for tuning results.
#> ℹ To predict with the optimal model configuration from tuning results,  ensure that the 
#>   tuning result was generated with the control option `save_workflow = TRUE` and then run:
#>     ames_mod <- fit_best(ames_grid_search)
#>     predict(ames_mod)
#> ℹ To collect predictions from tuning results, ensure that `ames_grid_search` 
#>   was generated with the control option `save_pred = TRUE` and run
#>   `collect_predictions(ames_grid_search)`.
```


hfrick (approved):
Very nice! I think this version is better than the alternative with `deparse(substitute())` on the input because we have the requirement on how the object was generated in there. Having seemingly executable code presented to you to then find out that you still need to do something else before that is disappointing at best, and confusing at worst. (Great of course if the object was already fit that way but arguing for the defensive version here.)

PR #689, opened 2023-06-10
Pivot metrics

topepo:
A function to collect metrics then `pivot_wider()` on them so that metrics are in columns.


simonpcouch (commented):
Been sitting with this and wanted to open a discussion before reviewing any technical bits: would this be better situated as an argument to `collect_metrics()`?

My hesitancy with `pivot_metrics()` is that the name currently suggests to me that its input is the output of `collect_metrics()` or `metrics()`, and the only alternatives I can think of for a name change seem ambiguous with `collect_metrics()`. I would also make the case this the name of this function should start with `collect_*()`, as this feels like a collection action and that naming would better support tab-completion. This function does feel more to me like an alternative usage of `collect_metrics()` than a different function altogether.

Transitioning this to a `pivot` argument with logical values, or `type = c("long", "wide")`, would also mean we wouldn't have to maintain a separate usage of this column identification boilerplate:

https://github.com/tidymodels/tune/blob/dde5d354a3125c4c1c84fef5b3bb4534d8c528a3/R/collect.R#L485-L495

simonpcouch:
Closing in favor of #839. :)

PR #702, opened 2023-07-17
Fix regression from fea8b02

topepo:
An unintended regression caused by fea8b02 . Test cases for survival models are in PRs, so there was very little chance of catching this. 

We might consider adding at least one live test case but that means adding survival to Suggests (which causes a bunch of other issues).


simonpcouch:
This is a duplicate of #700/#699. Would you prefer I close those in favor of this PR?

PR #712, opened 2023-08-08
Make testing more quiet

EmilHvitfeldt:
Not the biggest deal, but it removes a little bit of noise when running tests


hfrick (approved):
🧹

Issue #715, opened 2023-09-01
Extracting with `tune_bayes()` results in the same `.config` entry for each iteration


MasterLuke84:
Hello,

when using bayes tuning and extracting with `tune::extract_fit_engine` (s. reprex: the oob error from a `ranger` model),
the corresponding `.config`-column of `.extracts` contains always the value `Preprocessor1_Model1` for each iteration instead of `Iter1, Iter2, ...` (compare the `.config`-column of `.metrics`). 

``` r
library(magrittr)
library(yardstick)

set.seed(1234)

d <- mtcars %>% tibble::as_tibble()

prep_rec <- recipes::recipe(mpg ~ . ,data = d) %>% 
  recipes::prep()


rsmpl <- rsample::bootstraps(data = d, times = 2)  


mod_spec <- parsnip::rand_forest() %>%
  parsnip::set_engine("ranger") %>%
  parsnip::set_mode("regression") %>% 
  parsnip::set_args(mtry = 2, trees = tune())


param_set <- hardhat::extract_parameter_set_dials(mod_spec)

extract_oob <- function(x) { 
  tune::extract_fit_engine(x)$prediction.error
}

res <- tune::tune_bayes(object       = mod_spec, 
                        preprocessor = formula(prep_rec),
                        resamples    = rsmpl, 
                        metrics      = metric_set(rmse),
                        iter         = 3, 
                        param_info   = param_set, 
                        control      = tune::control_bayes(seed      = 1234, 
                                                           save_pred = TRUE, 
                                                           extract   = extract_oob))

# within .metrics the.config distingushes among initial combinations 
# (Preprocessor1_XXX) and iterations
res %>% dplyr::select(id, .metrics, .iter) %>% tidyr::unnest(.metrics)
#> # A tibble: 16 × 7
#>    id         trees .metric .estimator .estimate .config              .iter
#>    <chr>      <int> <chr>   <chr>          <dbl> <chr>                <int>
#>  1 Bootstrap1   234 rmse    standard        2.43 Preprocessor1_Model1     0
#>  2 Bootstrap1   884 rmse    standard        2.40 Preprocessor1_Model2     0
#>  3 Bootstrap1  1902 rmse    standard        2.34 Preprocessor1_Model3     0
#>  4 Bootstrap1   572 rmse    standard        2.31 Preprocessor1_Model4     0
#>  5 Bootstrap1  1284 rmse    standard        2.40 Preprocessor1_Model5     0
[...]

PR #717, opened 2023-09-07
regenerate example objects

simonpcouch:
In order to add tests for a fix for #715, I needed to regenerate the example test objects, but there were a few unrelated failures that needed test changes. I figured I'd PR the example object and test changes using `main` tune first and then follow up with only the needed changes for #715. :)

A whole lot of:

```
→ A | warning: prediction from a rank-deficient fit may be misleading
→ B | warning: some 'x' values beyond boundary knots may cause ill-conditioned bases
```

when regenerating the test objects.🥴


simonpcouch (commented):
It looks like these were [last updated in Oct 2020](https://github.com/tidymodels/tune/commits/0c1e5fa280e2ee96ed8df69a6ee6f81e0f3368bb/tests/testthat/test_objects.RData?browsing_rename_history=true&new_path=tests/testthat/data/test_objects.RData&original_branch=main).

hfrick (approved):
Looks all good to me. Thanks for attaching all the receipts for the changes! And I am GLAD these finally get updated 🥳

PR #725, opened 2023-09-27
simplify s3 registration in `.onLoad()`

### LINKED ISSUE #693: Simplify the S3 registration in `.onLoad()`
We can simplify the contents of `.onLoad()` because ggplot2, rsample, and dplyr are all Imports of tune so we don't need to register those methods lazily, we can just do it the regular way.

_Originally posted by @hfrick in https://github.com/tidymodels/tune/pull/688#discussion_r1236723624_

simonpcouch:
Closes #693. :)


R/reexports.R
```diff
@@ -3,6 +3,10 @@
 #' @export
 dials::parameters
 
+#' @importFrom dplyr dplyr_reconstruct
+#' @export
```
hfrick, line 7: I would not (re)export the generic here because I don't think anyone would (should) want to use `dplyr_reconstruct()` on its own, by just loading tune.

hfrick (approved):
🐬

R/reexports.R
simonpcouch: Heard! Since we're no longer re-exporting, I moved that `importFrom` from reexports.R to dplyr-compat.R.

PR #730, opened 2023-09-28
sort ids via vctrs, pass sorting to `pull_*()` helpers

### LINKED ISSUE #728: entries in `.notes` don't match the resample `id` they ought to
Entries in `.notes` are dropped in the wrong order in tuning results.

``` r
library(tidymodels)
  
mt_boots <- bootstraps(mtcars, 3, apparent = TRUE)
mt_boots
#> # Bootstrap sampling with apparent sample 
#> # A tibble: 4 × 2
#>   splits          id        
#>   <list>          <chr>     
#> 1 <split [32/14]> Bootstrap1
#> 2 <split [32/11]> Bootstrap2
#> 3 <split [32/13]> Bootstrap3
#> 4 <split [32/32]> Apparent

# induce the size zero yardstick error in Bootstrap1
mt_boots$splits[[1]]$out_id <- numeric(0)
mt_res <- fit_resamples(linear_reg(), mpg ~ ., mt_boots)
#> → A | warning: A correlation computation is required, but the inputs are size zero or one and the standard deviation cannot be computed. `NA` will be returned.
#> There were issues with some computations   A: x1
#> There were issues with some computations   A: x1
#> 

mt_res %>% select(id, .notes)
#> # A tibble: 4 × 2
#>   id         .notes          
#>   <chr>      <list>          
#> 1 Apparent   <tibble [1 × 3]>
#> 2 Bootstrap1 <tibble [0 × 3]>
#> 3 Bootstrap2 <tibble [0 × 3]>
#> 4 Bootstrap3 <tibble [0 × 3]>
```

<sup>Created on 2023-09-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

simonpcouch:
Closes #728.

### DIFF
--- a/R/grid_code_paths.R
+++ b/R/grid_code_paths.R
@@ -22,10 +22,14 @@ tune_grid_loop <- function(resamples,
     rng
   )
 
-  resamples <- pull_metrics(resamples, results, control)
-  resamples <- pull_notes(resamples, results, control)
-  resamples <- pull_extracts(resamples, results, control)
-  resamples <- pull_predictions(resamples, results, control)
+  # carry out arranging by id before extracting each element of results (#728)
+  resample_ids <- grep("^id", names(resamples), value = TRUE)
+  id_order <- vctrs::vec_order(resamples[resample_ids])
+
+  resamples <- pull_metrics(resamples, results, control, order = id_order)
+  resamples <- pull_notes(resamples, results, control, order = id_order)
+  resamples <- pull_extracts(resamples, results, control, order = id_order)
+  resamples <- pull_predictions(resamples, results, control, order = id_order)
   resamples <- pull_all_outcome_names(resamples, results)
 
   resamples
--- a/R/pull.R
+++ b/R/pull.R
@@ -10,7 +10,7 @@ extract_details <- function(object, extractor) {
 
 # Grab the new results, make sure that they align row-wise with the rsample
 # object and then bind columns
-pulley <- function(resamples, res, col, order = NULL) {
+pulley <- function(resamples, res, col, order) {
   if (all(purrr::map_lgl(res, inherits, "simpleError"))) {
     res <-
       resamples %>%
@@ -22,10 +22,6 @@ pulley <- function(resamples, res, col, order = NULL) {
   all_null <- all(purrr::map_lgl(res, is.null))
 
   id_cols <- grep("^id", names(resamples), value = TRUE)
-
-  if (is.null(order)) {
-    order <- vctrs::vec_order(resamples[id_cols])
-  }
 
   resamples <- vctrs::vec_slice(resamples, order)
 
--- a/R/pull.R
+++ b/R/pull.R
@@ -10,7 +10,7 @@ extract_details <- function(object, extractor) {
 
 # Grab the new results, make sure that they align row-wise with the rsample
 # object and then bind columns
-pulley <- function(resamples, res, col) {
+pulley <- function(resamples, res, col, order = NULL) {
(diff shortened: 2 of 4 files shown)


R/grid_code_paths.R
```diff
@@ -22,10 +22,14 @@ tune_grid_loop <- function(resamples,
     rng
   )
 
-  resamples <- pull_metrics(resamples, results, control)
-  resamples <- pull_notes(resamples, results, control)
-  resamples <- pull_extracts(resamples, results, control)
-  resamples <- pull_predictions(resamples, results, control)
+  # carry out arranging by id before extracting each element of results (#728)
```
simonpcouch, line 25: Previously this sorting happened independently for each of the `pull_*()` functions that call `pulley()`. For those that don't call `pulley()`, the ordering is wrong.

R/grid_code_paths.R
```diff
@@ -22,10 +22,14 @@ tune_grid_loop <- function(resamples,
...
+  resample_ids <- grep("^id", names(resamples), value = TRUE)
```
simonpcouch, line 26: This is the pattern used by `pulley()`.

R/grid_code_paths.R
```diff
@@ -22,10 +22,14 @@ tune_grid_loop <- function(resamples,
...
+  id_order <- vctrs::vec_order(resamples[resample_ids])
```
simonpcouch, line 27: I don't suspect this is any different than passing those columns to `arrange()` (some locale technicalities maybe?), but worth looking into.

R/pull.R
```diff
@@ -65,22 +71,22 @@ maybe_repair <- function(x) {
 }
 
 
-pull_metrics <- function(resamples, res, control) {
```
simonpcouch, line 68: The remainder of the changes in this file just newly pass `order` along to `pulley()`.

R/grid_code_paths.R
```diff
@@ -22,10 +22,14 @@ tune_grid_loop <- function(resamples,
...
+
+  resamples <- pull_metrics(resamples, results, control, order = id_order)
+  resamples <- pull_notes(resamples, results, control, order = id_order)
+  resamples <- pull_extracts(resamples, results, control, order = id_order)
+  resamples <- pull_predictions(resamples, results, control, order = id_order)
   resamples <- pull_all_outcome_names(resamples, results)
```
simonpcouch, line 33: Note that this function doesn't gain an `order` argument, as its output doesn't end up in tuning results.

R/grid_code_paths.R
EmilHvitfeldt: `arrange()` uses `vec_order_radix()` under the hood. but I don't think it should matter much here

R/pull.R
```diff
@@ -22,7 +22,13 @@ pulley <- function(resamples, res, col) {
   all_null <- all(purrr::map_lgl(res, is.null))
 
   id_cols <- grep("^id", names(resamples), value = TRUE)
-  resamples <- dplyr::arrange(resamples, !!!syms(id_cols))
+
+  if (is.null(order)) {
+    order <- vctrs::vec_order(resamples[id_cols])
+  }
+
```
EmilHvitfeldt, line 29: is this needed since you always pass in the order when `pulley()` is used?

EmilHvitfeldt (approved):
looking clean!

R/pull.R
simonpcouch: Thank you! Heard—[81749cc](https://github.com/tidymodels/tune/pull/730/commits/81749cc9a25c12c8b0970ec85531f1fb25872ab2).

PR #743, opened 2023-10-05
clarify hyperparameters associated with extracted objects

### LINKED ISSUE #724: hyperparameters associated to`extract`ed objects are incorrect
The hyperparameters attached to `extract` entries are incorrect for the models from the initial grid.

``` r
library(tidymodels)

set.seed(8825)
mt_knn_bo <-
  tune_bayes(
    workflow(mpg ~ ., nearest_neighbor("regression", neighbors = tune())),
    resamples = vfold_cv(mtcars, v = 5),
    iter = 3,
    control = control_bayes(extract = function(x) {list()})
  )

# same neighbors for all model configurations
mt_knn_bo %>%
  collect_extracts()
#> # A tibble: 30 × 5
#>    id    .iter neighbors .extracts  .config             
#>    <chr> <int>     <int> <list>     <chr>               
#>  1 Fold1     0        13 <list [0]> Preprocessor1_Model1
#>  2 Fold1     0        13 <list [0]> Preprocessor1_Model2
#>  3 Fold1     0        13 <list [0]> Preprocessor1_Model3
#>  4 Fold2     0        13 <list [0]> Preprocessor1_Model1
#>  5 Fold2     0        13 <list [0]> Preprocessor1_Model2
#>  6 Fold2     0        13 <list [0]> Preprocessor1_Model3
#>  7 Fold3     0        13 <list [0]> Preprocessor1_Model1
#>  8 Fold3     0        13 <list [0]> Preprocessor1_Model2
#>  9 Fold3     0        13 <list [0]> Preprocessor1_Model3
#> 10 Fold4     0        13 <list [0]> Preprocessor1_Model1
#> # ℹ 20 more rows

# though metrics shows neighbors did vary for these .configs
mt_knn_bo %>%
  collect_metrics()
#> # A tibble: 12 × 8
#>    neighbors .metric .estimator  mean     n std_err .config              .iter
#>        <int> <chr>   <chr>      <dbl> <int>   <dbl> <chr>                <int>
#>  1         4 rmse    standard   2.82      5  0.463  Preprocessor1_Model1     0
#>  2         4 rsq     standard   0.824     5  0.0331 Preprocessor1_Model1     0
#>  3         9 rmse    standard   2.80      5  0.517  Preprocessor1_Model2     0
#>  4         9 rsq     standard   0.823     5  0.0269 Preprocessor1_Model2     0
#>  5        13 rmse    standard   2.87      5  0.569  Preprocessor1_Model3     0
#>  6        13 rsq     standard   0.814     5  0.0377 Preprocessor1_Model3     0
#>  7         8 rmse    standard   2.80      5  0.510  Iter1                    1
#>  8         8 rsq     standard   0.826     5  0.0231 Iter1                    1
#>  9         7 rmse    standard   2.80      5  0.504  Iter2                    2
#> 10         7 rsq     standard   0.827     5  0.0203 Iter2                    2
#> 11         1 rmse    standard   3.13      5  0.374  Iter3                    3
#> 12         1 rsq     standard   0.775     5  0.0677 Iter3                    3
```

<sup>Created on 2023-09-27 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Note that the issue shown for `mt_knn_bo` applies to the object in `test_objects.RData` by the same name.

### LINKED ISSUE #733: audit `extracts` submodels FIXME
https://github.com/tidymodels/tune/blob/637e923e7ac9104a0d17f0e87122e77f888837db/R/grid_code_paths.R#L395-L400

May be related to https://github.com/tidymodels/tune/issues/724.

simonpcouch:
Closes #724, closes #733.

I went with the approach labeled `1.` in https://github.com/tidymodels/tune/issues/724#issuecomment-1749068714, but am open to other options. :)

### DIFF
--- a/R/collect.R
+++ b/R/collect.R
@@ -52,9 +52,10 @@
 #' indicators, the location (preprocessor, model, etc.), type (error or warning),
 #' and the notes.
 #'
-#' [collect_extracts()] returns a tibble with columns for the resampling
-#' indicators, the location (preprocessor, model, etc.), and objects extracted
-#' from workflows via the `extract` argument to [control functions][control_grid()].
+#' [collect_extracts()] collects objects extracted from fitted workflows
+#' via the `extract` argument to [control functions][control_grid()]. The
+#' function returns a tibble with columns for the resampling
+#' indicators, the location (preprocessor, model, etc.), and extracted objects.
 #'
 #' @section Hyperparameters and extracted objects:
 #'
--- a/R/collect.R
+++ b/R/collect.R
@@ -64,11 +64,11 @@
 #' However, this means that if a function was supplied to a
 #' [control function's][control_grid()] `extract` argument, tune can only
 #' execute that extraction on the one model that was fitted. As a result,
-#' in `collect_extracts()` output, tune opts to associate the hyperparameter
-#' combination used to fit the workflow that results in the extracted object
-#' with the extracted object, rather than with the `.config` entry that
-#' hyperparameter combination is usually associated with. In output, this
-#' appears like a hyperparameter entry is recycled across many `.config`
+#' in the `collect_extracts()` output, tune opts to associate the 
+#' extracted objects with the hyperparameter combination used to 
+#' fit that one model workflow, rather than the hyperparameter 
+#' combination of a submodel. In the output, this appears like 
+#' a hyperparameter entry is recycled across many `.config`
 #' entries---this is intentional.
 #'
 #' See \url{https://parsnip.tidymodels.org/articles/Submodels.html} to learn
--- a/R/collect.R
+++ b/R/collect.R
@@ -55,6 +55,24 @@
 #' [collect_extracts()] returns a tibble with columns for the resampling
 #' indicators, the location (preprocessor, model, etc.), and objects extracted
 #' from workflows via the `extract` argument to [control functions][control_grid()].
+#'
+#' @section Extracting with submodels:
+#'
+#' When making use of submodels, tune can generate predictions and calculate
+#' metrics for multiple model `.config`urations using only one model fit.
+#' However, this means that if a function was supplied to a
+#' [control function's][control_grid()] `extract` argument, tune can only
+#' execute that extraction on the one submodel that was fitted. As a result,
+#' in `collect_extracts()` output, tune opts to associate the hyperparameter
+#' combination used to fit the workflow that results in the extracted object
+#' with the extracted object, rather than with the `.config` entry that
+#' hyperparameter combination is usually associated with. In output, this
+#' appears like a hyperparameter entry is recycled across many `.config`
+#' entries---this is intentional.
+#'
+#' See \url{https://parsnip.tidymodels.org/articles/Submodels.html} to learn
+#' more about submodels.
+#'
 #' @examplesIf tune:::should_run_examples(suggests = "kknn")
 #' data("example_ames_knn")
 #' # The parameters for the model:
--- a/R/collect.R
+++ b/R/collect.R
@@ -56,7 +56,7 @@
 #' indicators, the location (preprocessor, model, etc.), and objects extracted
 #' from workflows via the `extract` argument to [control functions][control_grid()].
 #'
-#' @section Extracting with submodels:
+#' @section Hyperparameters and extracted objects:
 #'
 #' When making use of submodels, tune can generate predictions and calculate
 #' metrics for multiple model `.config`urations using only one model fit.
--- a/R/collect.R
+++ b/R/collect.R
@@ -63,7 +63,7 @@
 #' metrics for multiple model `.config`urations using only one model fit.
 #' However, this means that if a function was supplied to a
 #' [control function's][control_grid()] `extract` argument, tune can only
-#' execute that extraction on the one submodel that was fitted. As a result,
+#' execute that extraction on the one model that was fitted. As a result,
 #' in `collect_extracts()` output, tune opts to associate the hyperparameter
 #' combination used to fit the workflow that results in the extracted object
 #' with the extracted object, rather than with the `.config` entry that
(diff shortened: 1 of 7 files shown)


R/collect.R
```diff
@@ -55,6 +55,24 @@
...
 #' indicators, the location (preprocessor, model, etc.), and objects extracted
 #' from workflows via the `extract` argument to [control functions][control_grid()].
+#'
+#' @section Extracting with submodels:
+#'
+#' When making use of submodels, tune can generate predictions and calculate
+#' metrics for multiple model `.config`urations using only one model fit.
+#' However, this means that if a function was supplied to a
+#' [control function's][control_grid()] `extract` argument, tune can only
+#' execute that extraction on the one submodel that was fitted. As a result,
```
hfrick, line 65:
```suggestion
#' execute that extraction on the one model that was fitted. As a result,
```

R/collect.R
```diff
@@ -55,6 +55,24 @@
 #' [collect_extracts()] returns a tibble with columns for the resampling
```
hfrick, line 55: I think this part could put an emphasis on "gives you access to the fitted workflows" by, e.g., starting the content description with that.

R/collect.R
```diff
@@ -55,6 +55,24 @@
...
 #' indicators, the location (preprocessor, model, etc.), and objects extracted
 #' from workflows via the `extract` argument to [control functions][control_grid()].
+#'
+#' @section Extracting with submodels:
```
hfrick, line 59: This section title requires readers to already know what the submodel trick is. Can you think of a way of rephrasing that for someone just being confused with about the behavior of `collect_extacts()` and trying to figure out what's going on?

R/collect.R
hfrick: maybe move this into the section below (which could resolve the comment I made about the section title)?

R/collect.R
hfrick, line 65: I would avoid calling the fitted model a submodel. That conflicts with my understanding of "sub". 
I'd probably refer to the _fitted model_ and use _submodel_  for the unfitted model configs.

hfrick (approved):
Alright, here's a short summary of what we tossed around in our conversation just now: 

`collect metrics()` gives you the metrics based on the predictions (obviously) - and those are based on what was requested (here: different values for the number of neighbors). 
`collect_extracts()` promises to give you access to what was actually fitted (we should probably make that clearer in its description, too). 

We discuss some options that we discarded pretty quickly:
- Putting some cosmetics on: in the output of `collect_extract()`, we could make the content of `neighbors` reflect _what was requested_ to match the content of `.config` but that would not really resolve the tension arising from what was _requested_ not being what was _fitted_. It would just kick the point of realization about that difference a little further down the line by obfuscating it a little more. Bad idea.
- (Trying to) Hide the submodel trick: we could lean into "what was requested is the only thing the user sees" and try to extract the relevant submodel, aka the requested model instead of the fitted model. That is a lot of work and/or we litter it with bugs. Bad idea.
- Don't use the submodel trick: that would mean the difference that's causing us to think about all of this goes away. Cost: it'll be slooooow. Bad idea.

One alternative that was more interesting:
We also discussed making `collect_extracts()` return _only_ what was fitted, so remove rows with requested configs that we did not need to fit. That leans fully into the idea of `collect_extracts()` being about what was actually _fitted_. But what anchors all the results is what was _requested_, and we want users to easily be able to connect all outputs to that. Thus, we use that as the thing that dictates the format, even when it here means duplicating some entries. 

Conclusion: we decided to go with the current behavior and aim to document it better. I'm approving to signal that but think that it's worth considering wordsmithing the proposed documentation a bit more to reflect some more of the clarity we gained in our discussion. (I would but my brain is currently doing a fun overflow 😝 )

R/collect.R
simonpcouch: Heard. Attempted to resolve in [54f5074](https://github.com/tidymodels/tune/pull/743/commits/54f5074432d43a8e99c78a1563f465f001dbe9bf), will leave to you to decide there.

R/collect.R
simonpcouch: Attempted to resolve in [21932aa](https://github.com/tidymodels/tune/pull/743/commits/21932aaa897dc4941f660d3f1a41d219c874caef). I opted not to move this into the section below since that section is reused in the control docs, but will hear you out if you think that's how we ought to do it!

R/collect.R
simonpcouch: Heard! [39674de](https://github.com/tidymodels/tune/pull/743/commits/39674dea0213a9660988d9d964c258674568dede).

R/collect.R
```diff
@@ -52,9 +52,28 @@
...
+#' metrics for multiple model `.config`urations using only one model fit.
+#' However, this means that if a function was supplied to a
+#' [control function's][control_grid()] `extract` argument, tune can only
+#' execute that extraction on the one model that was fitted. As a result,
+#' in `collect_extracts()` output, tune opts to associate the hyperparameter
+#' combination used to fit the workflow that results in the extracted object
+#' with the extracted object, rather than with the `.config` entry that
+#' hyperparameter combination is usually associated with. In output, this
+#' appears like a hyperparameter entry is recycled across many `.config`
+#' entries---this is intentional.
```
hfrick, line 72:
```suggestion
#' execute that extraction on the one model that was fitted. As a result,
#' in the `collect_extracts()` output, tune opts to associate the 
#' extracted objects with the hyperparameter combination used to 
#' fit that one model workflow, rather than the hyperparameter 
#' combination of a submodel. In the output, this appears like 
#' a hyperparameter entry is recycled across many `.config`
#' entries---this is intentional.
```

hfrick (commented):
Nice changes! One round of wordsmithing for your consideration :D

R/collect.R
simonpcouch: I dig it. :)

PR #749, opened 2023-11-06
remove rsample `Remotes` entry

simonpcouch:
The `Remotes` entry for rsample was added four months ago:

https://github.com/tidymodels/tune/blob/3f82cb2a91b93d30f1fa48c64f83d4a5c23447f6/DESCRIPTION#L54-L55

and the 1.2.0 rsample release happened in the meantime. I believe we can drop this ref?


hfrick (approved):
Yes :) The remote was there for the `int_pctl()` generic (method added in https://github.com/tidymodels/tune/pull/688) which hit CRAN with rsample 1.2.0

Issue #759, opened 2023-11-20
augment.tune_results() doesn't return tibbles


EmilHvitfeldt:
``` r
library(tidymodels)
library(censored)
#> Loading required package: survival

set.seed(1)
sim_dat <- prodlim::SimSurv(500) %>%
  mutate(event_time = Surv(time, event)) %>%
  select(event_time, X1, X2)

set.seed(2)
split <- initial_split(sim_dat)
sim_tr <- training(split)
sim_te <- testing(split)
sim_rs <- vfold_cv(sim_tr)

time_points <- c(10, 1, 5, 15)

mod_spec <-
  proportional_hazards(penalty = tune(), mixture = 1) %>%
  set_engine("glmnet") %>%
  set_mode("censored regression")

grid <- tibble(penalty = 10^c(-4, -2, -1))

gctrl <- control_grid(save_pred = TRUE)

# Grid search with a mixture of metrics --------------------------------------

mix_mtrc <- metric_set(brier_survival, brier_survival_integrated, concordance_survival)

set.seed(2193)
tune_res <-
  mod_spec %>%
  tune_grid(
    event_time ~ X1 + X2,
    resamples = sim_rs,
    grid = grid,
    metrics = mix_mtrc,
    eval_time = time_points,
    control = gctrl
  )

augment(tune_res) |> class()
#> Warning: No evaluation time was set; a value of 5 was used.
#> [1] "data.frame"
```

<sup>Created on 2023-11-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #760, opened 2023-11-20
make augment methods always return tibbles

EmilHvitfeldt:
To close https://github.com/tidymodels/tune/issues/759

Original issue happened when the fitting data was a data.frame.


topepo (approved):

simonpcouch (approved):
Thank you. :)

PR #767, opened 2023-11-28
add eval_time as an attribute

topepo:
For survival models, we'll need the ordered evaluation times that were declared by `fit_resamples()`, `tune_*()`, and `last_fit()`. 

This adds an attribute and an accessor function for the `eval_time` argument. 

Similar changes need to occur in finetune too. 

Survival specific tests will be in extratests


EmilHvitfeldt (approved):
looking good. went search around for additional places where the change could be done. and didn't find any

Issue #775, opened 2023-12-02
renew cached objects for testing


topepo:
Scales 1.3.0 made a change in the object class of its transformers. Their new class ("transformers") is not compatible with the old class (see below). We need to recreate these with the new version of scales. 

``` r
library(tidymodels)
library(testthat)
#> 
#> Attaching package: 'testthat'
#> The following object is masked from 'package:tidyr':
#> 
#>     matches
#> The following object is masked from 'package:rsample':
#> 
#>     matches
#> The following object is masked from 'package:purrr':
#> 
#>     is_null
#> The following object is masked from 'package:dplyr':
#> 
#>     matches
library(scales)
```

``` r

tidymodels_prefer()
theme_set(theme_bw())
options(pillar.advice = FALSE, pillar.min_title_chars = Inf)
```

``` r
# uses old format transformers (pre scales v 1.3.0)

# Get current cached test object
temp_loc <- tempfile()
download.file(
  "https://github.com/tidymodels/tune/raw/main/tests/testthat/data/svm_reg_results.rds", 
  temp_loc)

svm_reg_results <- readRDS(temp_loc)
autoplot(svm_reg_results)
#> Error in `as.trans()`:
#> ! `trans` must be a character vector or a transformer object
#> Backtrace:
#>     ▆
#>  1. ├─ggplot2::autoplot(svm_reg_results)
#>  2. └─tune:::autoplot.tune_results(svm_reg_results)
#>  3.   └─tune:::plot_regular_grid(...) at tune/R/plots.R:106:11
#>  4.     └─ggplot2::scale_x_continuous(trans = trans) at tune/R/plots.R:634:5
#>  5.       └─ggplot2::continuous_scale(...)
#>  6.         └─scales::as.trans(trans)
#>  7.           └─cli::cli_abort(...)
#>  8.             └─rlang::abort(...)

# The dials objects are in the pre-made parameter sets 

dials_params <- tune:::get_param_object(svm_reg_results)
dials_params
#> Collection of 3 parameters for tuning
#> 
#>    identifier         type    object
#>          cost         cost nparam[+]
#>          %^*#       degree nparam[+]
#>  scale_factor scale_factor nparam[+]

scale_trans_obj <- dials_params$object[[3]]$trans

# with new scales: 
class(scale_trans_obj)
#> [1] "trans"
is.trans(scale_trans_obj)
#> [1] FALSE
as.trans(scale_trans_obj)
#> Error in `as.trans()`:
[...]

PR #776, opened 2023-12-02
New dials objects for #775

topepo:
#775 shows that we need new tune objects created with scales 1.3.0


topepo:
@hfrick I don't know if this affects any of your dials code or tests. We may want to bump the dials requirement for scales.

topepo:
Closes #775

PR #777, opened 2023-12-04
Refactor choose_metric

topepo:
Uses new code from #768. This only changes the metric code; evaluation times will be the next PR. 

Generally, the error messages are improved: 

old: (error) "Please check the value of `metric`.
new (warn): "'WAT' was not in the metric set. Please choose from: 'rmse', 'rsq'"

old (error) "Please specify a single character value for `metric`."
new (warn) "2 metrics were given; 'rmse' will be used"

This PR also gets rid of the old checks from the second version of tune: "The `maximize` argument is no longer needed."

Updated PR sequence: 

1. #768
1. (you are here) functions to validate the chosen metric with updates to `show_best()` and `select_best()`. 
1. functions to validate the chosen evaluation time with updates to `show_best()` and `select_best()` (see #766). 
1. update finetune to use the change
1. add additional functions that return (potentially) multiple evaluation times (eg `autoplot()`)
1. refactor metric input checking code in the `tune_*` functions, `fit_resamples()`. etc. 
1. write an article on tidymodels.org about this.

PR #768, opened 2023-11-28
Tools for selecting a default evaluation time

topepo:
`show_best()` will pick an evaluation time for a dynamic metric when none is given. 

Previously, we would find what was in the data and select a time that was close to the median time. This was fine but inconsistent with other parts of tidymodels that do similar operations. For example, tune_bayes has to have a metric to optimize on so it uses the first metric in the metric set and, if needed, the first evaluation time given to the function. 

This PR adds a few exported but internal helper functions (primarily `first_eval_time()`) as the canonical tools for these selections. 

This is one of a sequence of PRs:

1. you are here
1. functions to validate the chosen metric and evaluation times
1. refactor the internals of `show_best()` to be more modular with these tools
1. propagate these changes to other functions that either use `show_best()` or do similar computations (see #766)
1. update finetune to use the change
1. add additional functions that return (potentially) multiple evaluation times (eg `autoplot()`)
1. write an article on tidymodels.org about this.

Issue #766, opened 2023-11-21
When we need to default to a single value for `eval_time`


hfrick:
[Using this issue as the epic] 

Update (moved up here from https://github.com/tidymodels/tune/issues/766#issuecomment-1893596462 for better visibility)

- [x] `show_best()`: majority done, polishing required in https://github.com/tidymodels/tune/issues/781 and https://github.com/tidymodels/finetune/issues/89
- [x] `select_*()`: currently in sync wit `show_best()` by either calling it directly (`select_best()`) or calling `choose_eval_time()` like `show_best()` (the other `select_*()` functions)
- [x] `autoplot()`: needs to switch to using the first one https://github.com/tidymodels/tune/issues/808
- [x] `int_pctl()`: needs to switch to using the first one, plus something is broken now https://github.com/tidymodels/tune/issues/809
- [x] `augment()`: should not warn but use the first one quietly https://github.com/tidymodels/tune/issues/810
- [x] `fit_best()`: done in https://github.com/tidymodels/tune/pull/797

----

We allow `eval_time` to be a vector of time point for the tuning functions, also because the integrated survival metrics need multiple evaluation time points.

However, if we are optimizing for a dynamic metric (rather than integrated, or static) we are using the _first_ value of `eval_time` as the time point to calculate the metric at. 

We have a couple of other situations where we need to pick a single value for `eval_time` and we should do that consistently. The suggestion is to follow the logic of the tuning functions and pick the first value.

This affects the following functions if `eval_time = NULL`
- `show_best()`
- `select_*()`
- `autoplot()`
- `int_pctl()`

This also affects functions making use of `select_best()` in case the model choice is left unspecified
- `augment()`: no explict argument for `eval_time`. We should make a choice for `eval_time` like we do for the metric. That choice should be consistent with how we choose elsewhere.
- `fit_best()` also does not have an `eval_time` argument but does have a `metric` arg which defaults to the first metric if `NULL`

Note the inconsistency between `augment()` and `fit_best()` in terms of metrics argument here. 

``` r
library(tidymodels)
library(censored)
#> Loading required package: survival

set.seed(1)
sim_dat <- prodlim::SimSurv(500) %>%
  mutate(event_time = Surv(time, event)) %>%
  select(event_time, X1, X2) %>%
  as_tibble()

set.seed(2)
sim_rs <- vfold_cv(sim_dat)

time_points <- c(10, 1, 5, 15)

mod_spec <-
[...]

Issue #704, opened 2023-07-20
throw warning when eval_time isn't needed


topepo:
These functions should warn (but not error) when `eval_time` is _not_ `NULL` in the following cases:

For  non-survival models and for survival models with a static metric
- [x] `fit_resamples()` 
- [x] `tune_grid()` 
- [x] `tune_bayes()` 
- [x] `tune_race_anova()` works but https://github.com/tidymodels/finetune/issues/91
- [x] `tune_race_win_loss()` https://github.com/tidymodels/finetune/issues/92
- [x] `tune_sim_anneal()`

For non-survival models and for survival models _without_ a dynamic metric
- [x] `show_best()`and `select_best()` work but message text is not as closely aligned as it could be https://github.com/tidymodels/tune/issues/811
- [x] `select_by_one_std_err()`and `select_by_pct_loss()` use `choose_eval_time()`, thus same comment as `select_best()`
- [x] `autoplot()` #706
- [x] `int_pctl()`
- [x] `fit_best()`

And bringing it all together: https://github.com/tidymodels/tune/issues/811

And the exception, for consistency within the function (with how the choice of metric is treated)
- [x] `augment()` should not take an eval_time arg 

----

`show_best()` should throw a warning if `eval_time` is used but the metric selected is not dynamic. See last line: 

``` r
library(tidymodels)
library(censored)
#> Loading required package: survival
```

``` r

tidymodels_prefer()
theme_set(theme_bw())
options(pillar.advice = FALSE, pillar.min_title_chars = Inf)
```

``` r


data("mlc_churn")

mlc_churn <-
  mlc_churn %>%
  mutate(
    churned = ifelse(churn == "yes", 1, 0),
    event_time = Surv(account_length, churned)
  ) %>%
  select(event_time, account_length, voice_mail_plan) %>%
  slice(1:500)

set.seed(6941)
churn_split <- initial_split(mlc_churn, prop = 4/5)
churn_tr <- training(churn_split)
churn_te <- testing(churn_split)
churn_rs <- bootstraps(churn_tr, times = 2)
```

``` r

sr_tune_spec <- survival_reg(dist = tune())
eval_times <- c(10, 100, 150)
event_metrics <- metric_set(brier_survival, brier_survival_integrated,
                            concordance_survival, roc_auc_survival)
sr_grid <- tibble(dist = c("loglogistic", "lognormal"))
```

``` r

sr_tune_res <-
  sr_tune_spec %>%
  tune_grid(
[...]

PR #778, opened 2023-12-05
Refactored functions to select single evaluation times

topepo:
Primarily exports `choose_eval_time()` for use in select/show best functions. 

More changes will be coming for `autoplot()` and similar. 

Places where `select_best()` is used: [`augment.tune_results()`](https://github.com/tidymodels/tune/blob/1b67e42b3c3aee416fca468315fcd6a6aabdc676/R/augment.R#L50), [`fit_best()`](https://github.com/tidymodels/tune/blob/1b67e42b3c3aee416fca468315fcd6a6aabdc676/R/fit_best.R#L103). These will need more work since they do not have `eval_time` arguments. 

There is also still some legacy code that overlaps with this and previous changes (e.g. `middle_eval_time()`). These will be removed in a subsequent PR. 

1. #768
1. #777 
1. (you are here) functions to validate the chosen evaluation time with updates to `show_best()` and `select_best()` (see #766). 
1. update finetune to use the change
1. add additional functions that return (potentially) multiple evaluation times (eg `autoplot()`)
1. refactor metric input checking code in the `tune_*` functions, `fit_resamples()`. etc. 
1. write an article on tidymodels.org about this.


R/metric-selection.R
```diff
@@ -99,8 +149,39 @@ first_eval_time <- function(mtr_set, metric = NULL, eval_time = NULL) {
...
+    summary_res <- summary_res[summary_res$.eval_time == eval_time, ]
+  }
+  if (nrow(summary_res) == 0) {
+    cli::cli_abort("No results are available. Please use {.code collect_metrics()} to see if there were any issues.")
+  }
+
+  summary_res
+}
+
+# TODO will be removed shortly
```
topepo, line 178: I can't wait to get rid of this.

R/select_best.R
```diff
@@ -79,23 +79,9 @@ show_best.tune_results <- function(x, metric = NULL, n = 5, eval_time = NULL, ..
   metric_info <- choose_metric(x, metric)
   metric <- metric_info$metric
 
-  summary_res <- estimate_tune_results(x)
+  eval_time <- choose_eval_time(x, metric, eval_time = eval_time)
 
-  # TODO
```
topepo, line 84: A fair amount of collateral refactoring happens here to more modularize the selection/filtering code.

R/metric-selection.R
```diff
@@ -58,6 +67,47 @@ contains_survival_metric <- function(mtr_info) {
...
 
+#' @rdname choose_metric
+#' @export
+choose_eval_time <- function(x, metric, eval_time = NULL, call = rlang::caller_env()) {
+  mtr_set <- .get_tune_metrics(x)
+  mtr_info <- tibble::as_tibble(mtr_set)
+
+  if (!contains_survival_metric(mtr_info)) {
+    return(NULL)
+  }
```
hfrick, line 78: With this, we don't warn when a user supplies `eval_time` together with a non-survival metric but want to ( #704) - should that be happening in this PR?

R/metric-selection.R
```diff
@@ -58,6 +67,47 @@ contains_survival_metric <- function(mtr_info) {
...
+}
+
+is_dyn <- function(mtr_set, metric) {
+  mtr_info <- tibble::as_tibble(mtr_set)
+  mtr_cls <- mtr_info$class[mtr_info$metric == metric]
+  mtr_cls  == "dynamic_survival_metric"
+}
+
+check_right_eval_time <- function(x, eval_time, call = rlang::caller_env()) {
+  given_times <- .get_tune_eval_times(x)
```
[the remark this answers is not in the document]
hfrick: Can we rename the `check_right_*()` functions into something that says what "right" means here? Maybe `check_*_in_tune_results()`?

R/metric-selection.R
```diff
@@ -58,6 +67,47 @@ contains_survival_metric <- function(mtr_info) {
...
+  mtr_set <- .get_tune_metrics(x)
+  mtr_info <- tibble::as_tibble(mtr_set)
+
+  if (!contains_survival_metric(mtr_info)) {
+    return(NULL)
+  }
+
+  # If we need an eval time, set it to the possible values so that
+  # we can choose the first value
+  if (is_dyn(mtr_set, metric) & is.null(eval_time)) {
```
hfrick, line 82:
```suggestion
  if (is_dyn(mtr_set, metric) && is.null(eval_time)) {
```

hfrick (approved):
🚂

R/metric-selection.R
topepo: Why not `&&`?

tests/testthat/_snaps/select_best.md
```diff
@@ -213,3 +213,568 @@
...
 
+# show_best with survival models
+
+    Code
+      show_best(surv_res)
+    Condition
+      Warning in `show_best()`:
+      No value of `metric` was given; "brier_survival" will be used.
+      Warning:
+      4 evaluation times were specified; the first (10) will be used.
```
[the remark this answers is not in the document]
topepo: I deliberately used "specified" since it reflects that it is what the user supplied to the argument. I prefer that.

R/metric-selection.R
hfrick: I'm suggesting `&&` because it enforces a single TRUE/FALSE so that it errors on the comparison rather than the `if` statement it's in

``` r
rep(TRUE, 2) & rep(FALSE, 2)
#> [1] FALSE FALSE
rep(TRUE, 2) && rep(FALSE, 2)
#> Error in rep(TRUE, 2) && rep(FALSE, 2): 'length = 2' in coercion to 'logical(1)'
```

<sup>Created on 2023-12-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

tests/testthat/_snaps/select_best.md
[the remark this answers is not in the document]
hfrick: Ah, I see. What about "X evaluation times were specified during tuning; the first will be used"? The intent is to make it clear that it was indeed the user who specified that and when/where.

R/metric-selection.R
```diff
@@ -58,6 +67,47 @@ contains_survival_metric <- function(mtr_info) {
...
+  mtr_cls  == "dynamic_survival_metric"
+}
+
+check_right_eval_time <- function(x, eval_time, call = rlang::caller_env()) {
+  given_times <- .get_tune_eval_times(x)
+  if (!is.null(eval_time)) {
+    if (!any(eval_time == given_times)) {
+      num_times <- length(given_times)
+      print_time <- format(eval_time, digits = 3)
+      cli::cli_abort("Evaluation time {.val {print_time}} is not in the results.", call = call)
```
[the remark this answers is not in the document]
topepo: We might be using this in a slightly broader context than tuning (like simple resampling or last fit).

Issue #758, opened 2023-11-16
GP: X should be in range (0, 1)


topepo:
This should not issue a warning. Maybe the intervals should be inclusive? 

``` r
suppressPackageStartupMessages(library(tidymodels))
suppressPackageStartupMessages(library(censored))

set.seed(1)
sim_dat <- prodlim::SimSurv(500) %>%
  mutate(event_time = Surv(time, event)) %>%
  select(event_time, X1, X2)

set.seed(2)
split <- initial_split(sim_dat)
sim_tr <- training(split)
sim_te <- testing(split)
sim_rs <- vfold_cv(sim_tr, repeats = 3)

time_points <- c(10, 1, 5, 15)

mod_spec <-
  decision_tree(tree_depth = tune(), min_n = 4) %>%
  set_engine("partykit") %>%
  set_mode("censored regression")

grid <- tibble(tree_depth = c(1, 2, 5))

gctrl <- control_grid(save_pred = TRUE)
bctrl <- control_bayes(save_pred = TRUE)

# Bayes with static metrics --------------------------------------------------

stc_mtrc  <- metric_set(concordance_survival)

set.seed(2193)
init_grid_static_res <-
  mod_spec %>%
  tune_grid(
    event_time ~ X1 + X2,
    resamples = sim_rs,
    grid = grid,
    metrics = stc_mtrc,
    control = gctrl
  )

mod_spec %>% extract_parameter_set_dials() %>% pluck("object")
#> [[1]]
#> Tree Depth (quantitative)
#> Range: [1, 15]

set.seed(2193)
bayes_static_res <-
  mod_spec %>%
  tune_bayes(
    event_time ~ X1 + X2,
    resamples = sim_rs,
    iter = 1,
    metrics = stc_mtrc,
    control = bctrl,
    initial = init_grid_static_res
  )
#> → A | warning: X should be in range (0, 1)
#> There were issues with some computations   A: x1
#> There were issues with some computations   A: x1
#> 
```

<sup>Created on 2023-11-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

topepo:
It turns out that [the message](https://github.com/cran/GPfit/blob/master/R/GP_fit.R#L151C1-L153C6) is not accurate. It is complaining that  `(max(X) - min(X)) <= 0.5`. 

I don't think that this is an issue (in terms of the quality of the model). We could move to another package to avoid the warning. For example, this package has about the same results: 

``` r
[...]

PR #788, opened 2023-12-15
muffle "X should be in range (0, 1)" warning

EmilHvitfeldt:
This PR aims to close https://github.com/tidymodels/tune/issues/758.

A calling handler is used, and when then warning `"X should be in range (0, 1)"` is found, then it is muffled

### DIFF
--- a/R/tune_bayes.R
+++ b/R/tune_bayes.R
@@ -578,13 +578,23 @@ fit_gp <- function(dat, pset, metric, control, eval_time = NULL, ...) {
   }
 
   opts <- list(...)
+
+  withCallingHandlers({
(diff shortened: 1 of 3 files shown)


R/tune_bayes.R
```diff
@@ -578,13 +578,21 @@ fit_gp <- function(dat, pset, metric, control, eval_time = NULL, ...) {
   }
 
   opts <- list(...)
+
+  withCallingHandlers({
```
EmilHvitfeldt, line 582: i chose to do one call to `withCallingHandlers()` around the whole `if` instead of two because it felt better.

no strong opinion either way

PR #797, opened 2024-01-05
fit_best with survival

topepo:
We had not added an `eval_time` argument here. The change is simple since `select_best()` is doing all of the new work. 

Unit tests are in extratests.


R/fit_best.R
```diff
@@ -8,6 +8,9 @@
 #' [`save_workflow = TRUE`][tune::control_grid] should have been used.
 #' @param metric A character string (or `NULL`) for which metric to optimize. If
 #' `NULL`, the first metric is used.
+#' @param eval_time A numeric vector of time points where dynamic event time
+#' metrics should be chosen (e.g., the time-dependent ROC curve, etc). The
+#' values should be consistent with the values used to create `x`.
```
hfrick, line 13: Can you add what happens if `eval_time = NULL`?

R/fit_best.R
```diff
@@ -100,7 +104,7 @@ fit_best.tune_results <- function(x,
     if (is.null(metric)) {
       metric <- .get_tune_metric_names(x)[1]
```
hfrick, line 105: If `eval_time = NULL`, we want to choose the first one used in the `tune_results` object and document that, rather than pass it on directly to `show_best()` because that will issue a warning. This is to handle `eval_time` analogous to how we handle `metric` here for `fit_best()`.

https://github.com/tidymodels/tune/issues/766#issuecomment-1822971743

hfrick (commented):
Small change request! :)

R/fit_best.R
topepo: See changes: It will silently use the first given time for dynamic metrics and passes NULL otherwise.

hfrick (approved):

PR #798, opened 2024-01-05
Survival summarize predictions include tuning parameters and .iter

topepo:
These were some missing `.by` arguments and the columns were left out of the results. 

Also reorders the columns of `collect_predictions()` so that `starts_with(".pred")` columns are to the right (previously scattered, esp in survival results). 

Tests in extratests.


hfrick (commented):
I've left a whole bunch of comments on the corresponding test PR but those largely don't seem to stem from this PR. The one I would investigate before moving further with this PR is https://github.com/tidymodels/extratests/pull/158#discussion_r1445157478

Additionally, I don't think this actually closes #800 , I'm gonna edit the PR description to reflect that so that we don't accidentally close it.

PR #807, opened 2024-01-12
More case weight documentation

### LINKED ISSUE #805: more docs on case weights
We should have details that mention using a workflow and `add_case_weights()`. 

Having a section in Details on "Using Case Weights" could talk about

 - that you have to use a workflow
 - our special functions for those
 - `add_case_weights()`
 - how to determine if the model you are interested in can handle them. (`parsnip:::case_weights_allowed()` - maybe we should export that)

topepo:
Closes #805 

Uses a newly exported function from parsnip so there is a parsnip remote now. 

Will do the same to finetune functions next.

### DIFF
--- a/man-roxygen/case-weights.R
+++ b/man-roxygen/case-weights.R
@@ -6,7 +6,7 @@
 #' fitting.
 #'
 #' To know if your model is capable of using case weights, create a model spec
-#' and run `parsnip:::case_weights_allowed(model_spec)`.
+#' and test it using [parsnip::case_weights_allowed()].
 #'
 #' To use them, you will need a numeric column in your data set that has been
 #' passed through either [hardhat:: importance_weights()] or
--- a/man-roxygen/case-weights.R
+++ b/man-roxygen/case-weights.R
@@ -2,7 +2,7 @@
 #' Some models can utilize case weights during training. tidymodels currently
 #' supports two types of case weights: importance weights (doubles) and
 #' frequency weights (integers). Frequency weights are used during model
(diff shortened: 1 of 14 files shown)


man-roxygen/case-weights.R
```diff
@@ -0,0 +1,22 @@
+#' @section Case Weights:
+#' Some models can utilize case weights during training. tidymodels currently
+#' supports two types of case weights: importance weights (doubles) and
+#' frequency weights (integers). Frequency weights are used during model
+#' fitting and evaluation whereas Importance weights are only used during
```
hfrick, line 5:
```suggestion
#' fitting and evaluation whereas importance weights are only used during
```

man-roxygen/case-weights.R
```diff
@@ -0,0 +1,22 @@
...
+#' passed through either [hardhat:: importance_weights()] or
+#' [hardhat::frequency_weights()].
+#'
+#' For functions such as [fit_resamples()] and the `tune_*()` functions, the
+#' model must be contained inside of a [workflows::workflow()]. To declare that
+#' case weights are used, invoke [workflows::add_case_weights()] with the
+#' corresponding (unquoted) column name.
+#'
+#' From there, the packages will appropriately handle the weights during model
+#' fit and (if appropriate) performance estimation.
```
EmilHvitfeldt, line 21:
```suggestion
#' fitting and (if appropriate) performance estimation.
```

EmilHvitfeldt (approved):
Looking good. 

I think we should link it in `fit_best()` since we linking in `last_fit()`

PR #818, opened 2024-01-18
bug fixes for percentile intervals on survival metrics

topepo:
The initial implementation was only for purely dynamic metrics. 

Changes:

* Adds a `quietly` argument to `choose_eval_time()`
* Adds a function(`maybe_choose_eval_time()`) that runs `choose_eval_time()` across the metrics in a metric set to get defaults for all metrics. If an integrated metric is in the set, all of the original evaluation times (from the tune object) are returned. 
* Passed missing `eval_time` arguments through the missing bootstrap interval machinery. 
* Updated/fixed two existing tests for `int_pctl()`

Tests to come in extratests

_note_: the percentile intervals are calculated for each value of the `term` column (via a [`group_by()`](https://github.com/tidymodels/rsample/blob/main/R/bootci.R#L276). It would greatly simplify things here if we expanded that to include a few other columns: `.config`, `.iter`, and `.eval_time`. In fact, it might help users if we were to group by `term` and `starts_with(".")`. I'll start an issue for this in rsample.


topepo:
tidymodels/rsample#465 includes a proposal for some rsample changes. I don't want to force an rsample release so we can keep this PR as-is and, if the rsample changes are ok, update this code later.

R/int_pctl.R
```diff
@@ -137,29 +157,36 @@ boostrap_metrics_by_config <- function(config, seed, x, metrics, times, allow_pa
...
-  times <- dplyr::tibble(.eval_time = time)
-  x$.metrics <- purrr::map(x$.metrics, ~ dplyr::inner_join(.x, times, by = ".eval_time"))
-  rsample::int_pctl(x, .metrics) %>%
-    dplyr::mutate(.eval_time = time) %>%
-    dplyr::relocate(.eval_time, .after = term)
-}
+# tests in extratests
+# nocov start
+int_pctl_surv <- function(x, allow_par, alpha) {
+  `%op%` <- get_int_p_operator(allow_par)
```
hfrick, line 168: Do we still need this? Are you removing parallelization?

NAMESPACE
```diff
@@ -210,6 +210,7 @@ export(is_recipe)
 export(is_workflow)
 export(last_fit)
 export(load_pkgs)
+export(maybe_choose_eval_time)
```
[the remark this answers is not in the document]
topepo: Mostly future-proofing; I can see us needing this a lot inside of tune but also maybe finetune or future extensions. There's no harm in exporting and we've re-exported a fair number of functions that were originally internal.

R/int_pctl.R
topepo: Yes we do. This recomputed thousands of metrics and does add some computational efficiency.

hfrick: I thought so, I meant more "does this get picked up?" because I do see an explicit call to `foreach::foreach()` in the previous version but in the new version?

topepo: The changes remove the need for parallelization here; we do it at the previous level of computations.

Issue #820, opened 2024-01-19
Pass call from `select_best()` to `first_eval_time()`


hfrick:
``` r
library(tidymodels)
library(censored)
#> Loading required package: survival

lung_surv <- lung %>%
  dplyr::mutate(surv = Surv(time, status), .keep = "unused")

set.seed(2193)
tune_res <-
  proportional_hazards(penalty = tune(), engine = "glmnet") %>%
  tune_grid(
    surv ~ .,
    resamples = vfold_cv(lung_surv, 2),
    grid = 2,
    control = control_grid(save_pred = TRUE, save_workflow = TRUE),
    metrics = metric_set(brier_survival),
    eval_time = c(10, 20)
  )

# what's with the second call?
foo <- select_best(tune_res)
#> Warning in select_best(tune_res): No value of `metric` was given;
#> "brier_survival" will be used.
#> Warning: 2 evaluation times are available; the first (10) will be used.
```

<sup>Created on 2024-01-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #822, opened 2024-01-19
let `first_eval_time()` receive a `call`

### LINKED ISSUE #819: Pass call from `select_best()` to `show_best()`
``` r
library(tidymodels)
library(censored)
#> Loading required package: survival

lung_surv <- lung %>%
  dplyr::mutate(surv = Surv(time, status), .keep = "unused")

set.seed(2193)
tune_res <-
  proportional_hazards(penalty = tune(), engine = "glmnet") %>%
  tune_grid(
    surv ~ .,
    resamples = vfold_cv(lung_surv, 2),
    grid = 2,
    control = control_grid(save_pred = TRUE, save_workflow = TRUE),
    metrics = metric_set(concordance_survival, brier_survival_integrated, brier_survival),
    eval_time = c(10, 20)
  )
# show select_best() as call, not show_best()
foo <- select_best(tune_res, metric = "concordance_survival", eval_time = 0)
#> Warning in show_best(x, metric = metric, n = 1, eval_time = eval_time): An evaluation time is only required when a dynamic metric is selected (and
#> `eval_time` will thus be ignored).
```

<sup>Created on 2024-01-19 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

hfrick:
part of  #820 

This is passing the call from show_best() to first_eval_time(). Passing it the "first" leg from select_best() down will happen when we close #819 for passing the call from select_best() to show_best()

### DIFF
--- a/R/metric-selection.R
+++ b/R/metric-selection.R
@@ -7,6 +7,7 @@
 #' @param wflow A [workflows::workflow()].
 #' @param x An object with class `tune_results`.
 #' @param call The call to be displayed in warnings or errors.
+#' @inheritParams rlang::args_dots_empty
 #' @details
 #' These are developer-facing functions used to compute and validate choices
 #' for performance metrics. For survival analysis models, there are similar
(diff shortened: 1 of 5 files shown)


EmilHvitfeldt (commented):
Looking clean!

Issue #706, opened 2023-07-28
`autoplot()` should warn when used with non-needed `eval_time`


hfrick:
The `autoplot()` method for tune results based on an integrated survival metric should warn when used with an `eval_time` argument because the metric is integrated _over eval time_. 

``` r
library(tidymodels)
library(censored)
#> Loading required package: survival

set.seed(1)
sim_dat <- prodlim::SimSurv(500) %>%
  mutate(event_time = Surv(time, event)) %>%
  select(event_time, X1, X2)

sim_rs <- vfold_cv(sim_dat)

time_points <- c(10, 1, 5, 15)

mod_spec <-
  decision_tree(tree_depth = tune(), min_n = 4) %>%
  set_engine("partykit") %>%
  set_mode("censored regression")

sint_mtrc <- metric_set(brier_survival_integrated)

grid <- tibble(tree_depth = c(1, 2, 10))

set.seed(2193)
grid_integrated_res <-
  mod_spec %>%
  tune_grid(
    event_time ~ X1 + X2,
    resamples = sim_rs,
    grid = grid,
    metrics = sint_mtrc,
    eval_time = time_points
  )

# this should warn because it's an integrated metric
autoplot(grid_integrated_res, eval_time = 10)
```

<sup>Created on 2023-07-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

hfrick:
part of  #704

hfrick:
Comprehensive reprex for
- not a survival model
- survival model with static metric
- survival model with integrated metric

``` r
library(tidymodels)
library(censored)
#> Loading required package: survival
library(finetune)

lung_surv <- lung %>% 
  dplyr::mutate(surv = Surv(time, status), .keep = "unused")

# mode is not censored regression 
set.seed(2193)
tune_res <-
  linear_reg(penalty = tune(), engine = "glmnet") %>%
  tune_grid(
    mpg ~ .,
    resamples = vfold_cv(mtcars, 2),
    metrics = metric_set(rmse),
    grid = 2
  )
foo <- autoplot(tune_res, metric = "rmse", eval_time = 10)

# non-dynamic metric 
set.seed(2193)
tune_res <-
  proportional_hazards(penalty = tune(), engine = "glmnet") %>%
  tune_grid(
    surv ~ .,
    resamples = vfold_cv(lung_surv, 2),
    grid = 2,
    metrics = metric_set(concordance_survival, brier_survival_integrated),
[...]

PR #837, opened 2024-01-30
Better warning in autoplot() for unneeded evaluation time

topepo:
Closers #706 

I'll add more survival-specific tests in extratests


EmilHvitfeldt (approved):
looking good!

PR #839, opened 2024-01-31
add `collect_metrics()` argument to pivot output

simonpcouch:
A follow-up on #689.

Still needs tests for survival and finetuning, to come in extratests. Opening as draft as I haven't began writing those tests.

### DIFF
--- a/R/collect.R
+++ b/R/collect.R
@@ -11,6 +11,12 @@
 #'  used to filter the predicted values before processing. This tibble should
 #'  only have columns for each tuning parameter identifier (e.g. `"my_param"`
 #'  if `tune("my_param")` was used).
+#' @param type One of `"long"` (the default) or `"wide"`. When `type = "long"`,
+#'  output has columns `.metric` and one of `.estimate` or `mean`.
+#'  `.estimate`/`mean` gives the values for the `.metric`. When `type = "wide"`,
+#'  each metric has its own column and the `n` and `std_err` columns are removed,
+#'  if they exist.
+#'
 #' @param ... Not currently used.
 #' @return A tibble. The column names depend on the results and the mode of the
 #' model.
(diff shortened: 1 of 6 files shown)


R/collect.R
```diff
@@ -11,17 +11,26 @@
 #'  used to filter the predicted values before processing. This tibble should
 #'  only have columns for each tuning parameter identifier (e.g. `"my_param"`
 #'  if `tune("my_param")` was used).
+#' @param type One of `"long"` (the default) or `"wide"`. When `type = "long"`,
```
simonpcouch, line 14: I'm conflicted on the name and values for this argument. Very much open to suggestions. :)

R/collect.R
```diff
@@ -11,17 +11,26 @@
...
+#'  output has columns `.metric` and one of `.estimate` or `mean`.
+#'  `.estimate`/`mean` gives the values for the `.metric`. When `type = "wide"`,
+#'  each metric has its own column and the `n` and `std_err` columns are removed,
+#'  if they exist.
```
simonpcouch, line 18: I didn't have strong opinions here and the original implementation in #689 had the same behavior, so I did it this way, _but_ if we wanted we could have columns like `rmse`, `rmse_n`, `rmse_std_err`.

R/collect.R
```diff
@@ -445,19 +454,41 @@ collect_metrics.default <- function(x, ...) {
...
 #' @export
 #' @rdname collect_predictions
-collect_metrics.tune_results <- function(x, summarize = TRUE, ...) {
+collect_metrics.tune_results <- function(x, summarize = TRUE, type = c("long", "wide"), ...) {
+  rlang::arg_match0(type, values = c("long", "wide"))
+
   if (inherits(x, "last_fit")) {
-    return(x$.metrics[[1]])
+    res <- x$.metrics[[1]]
+  } else {
```
simonpcouch, line 462: Not stoked about this design pattern.😞 But wasn't clear to me that there's a super clean way to write this.

simonpcouch:
finetune has its own `collect_metrics()` methods, so this PR and its tests can function relatively independently of similar changes that would happen in finetune and a paired extratests PR.

simonpcouch:
Ah, I had forgotten that `collect_metrics.tune_race()` calls `NextMethod()`. Just added survival tests. :) tidymodels/extratests@abaa263

R/collect.R
topepo: I think excluding them is fine; we can add them later if someone asks for them.

R/collect.R
topepo: I think that a logical called `pivot_wider` is good. It implies that it is already long.

topepo (commented):
Looks great. Thanks!

R/collect.R
hfrick: Just to check, we decided that it's unlikely that we'll ever deviate from something binary? No other forms of "wide" metrics anticipated?

hfrick (approved):

topepo:
> Just to check, we decided that it's unlikely that we'll ever deviate from something binary? No other forms of "wide" metrics anticipated?

I've thought about it some more and I don't think that we will have more ways of pivoting (in tune, at least). 

characterize might. It is conceivable that characterize would have up to three options for pivoting but I think that its API would be consistent with the logical/binary argument ☝️

simonpcouch:
The only other form of "wide" or "long" I could imagine is one a la `pivot_longer(usual_output, cols = tuning_parameters)`, where we end up with a column "parameter" and a column "value." I'm not sure in what situations that'd be useful as a pattern.

PR #848, opened 2024-02-14
Harmonize docs for `eval_time`

### LINKED ISSUE #830: harmonize eval_time argument documentation
Right now we are manually documenting `eval_time` many different places. We could do a little more inheriting. 

https://github.com/search?q=repo:tidymodels/tune%20@param%20eval_time&type=code

Screenshot 2024-01-23 at 12 31 17 PM

hfrick:
closes #830

### DIFF
--- a/R/select_best.R
+++ b/R/select_best.R
@@ -32,7 +32,8 @@
 #' least one term is required for these two functions. See the examples below.
 #' @param eval_time A numeric vector of time points where dynamic event time
 #' metrics should be chosen (e.g., the time-dependent ROC curve, etc). The
-#' values should be consistent with the values used to create `x`.
+#' values should be consistent with the values used to create `x`. The `NULL` 
+#' default will automatically use the first evaluation time used by `x`.
 #' @param call The call to be shown in errors and warnings.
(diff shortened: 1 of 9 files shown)


R/select_best.R
```diff
@@ -32,7 +32,8 @@
 #' least one term is required for these two functions. See the examples below.
 #' @param eval_time A numeric vector of time points where dynamic event time
```
topepo, line 33: This is where I hate word wrapping. 

We should expect a single time point (or null) for these functions as well as `fit_best()`. (right?)

So how about: 

> A single numeric time point where dynamic event time metrics should be assessed (e.g., the time-dependent ROC curve, etc). The value should be consistent with the values used to create `x`. The `NULL` default will automatically use the first evaluation time used by `x`.

topepo (commented):
Should we add more details to the "censored regression" section in the details for `tune_grid()`?  That doesn't say much and the context assumes that we are using dynamic metrics (IMO): 

> Values of `eval_time` should be less than the largest observed event time in the training data. For many non-parametric models, the results beyond the largest time corresponding to an event are constant (or `NA`).

Maybe that is a different issue/PR.

PR #858, opened 2024-02-27
update snapshot for ggplot2 v3.5.0

hfrick:
Alternatively: remove the snapshot since it covers a warning emitted from ggplot2, not tune.


hfrick:
I see a hint of "dealing with NA" in the code for `CoordObsPred()` so I assume that test did have some particular intent and should be changed rather than removed. I'll put on my ruthlessly-removing-technical-debt hat some other time again 🙃

PR #861, opened 2024-02-28
Remove `eval_time` from `.estimate_metrics()`

### LINKED ISSUE #856: Remove `eval_time` argument from `.estimate_metrics()`
This looks like a leftover from the time when we thought we'd pass the evaluation time to the metric function. It's not being used so we should remove the argument.

hfrick:
closes #856

I went ahead and removed dead eval_time arguments up the call stack. It's still an argument to `int_pctl()` because that allows you to specify a subset the evaluation times used to generate the `tune_results` object. I've added a test in https://github.com/tidymodels/extratests/pull/206 to show that this feature still works after the changes here.

### DIFF
--- a/R/grid_performance.R
+++ b/R/grid_performance.R
@@ -59,7 +59,7 @@ metrics_info <- function(x) {
 #' @name tune-internal-functions
 #' @export
 .estimate_metrics <- function(dat, metric, param_names, outcome_name, event_level,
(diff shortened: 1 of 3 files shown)


EmilHvitfeldt (commented):
looking good!

EmilHvitfeldt (approved):
looking good!

PR #862, opened 2024-02-28
Remove duplicated code in `int_pctl()`

### LINKED ISSUE #855: Remove code duplication from `int_pctl.tune_results()`
In `int_pctl.tune_results()`, we have the same code block twice so we should most likely remove one of them. I am assuming the older one (added 7 months ago in https://github.com/tidymodels/tune/pull/688 vs last month in https://github.com/tidymodels/tune/pull/818) is the one to remove

https://github.com/tidymodels/tune/blob/9203a20ba4be712e6bab7542826b2df1ed9da53b/R/int_pctl.R#L101-L107

hfrick:
closes #855


simonpcouch (approved):
🧹

PR #909, opened 2024-05-31
pass `labels(split)` rather than `split` to internal functions

### LINKED ISSUE #893: pass only `labels(split)` to internal functions that just take `split` for its `labels()`
In `tune_grid_loop_iter()`, we pass `split` to several error handling functions that only use the split to generate informative error messages, i.e. with `labels(split)`. We should instead just write:

```r
split_label <- labels(split)
```

at the [top of the function](https://github.com/tidymodels/tune/blob/c6cf69f5ebd7d7c4af554b1a341ebc26f2f2cfbb/R/grid_code_paths.R#L337) and rewrite those helpers to just take the label so that it's clearer why that's being passed around. This will be especially helpful once a post-processor is plumbed through that further splits `split` to properly evaluate the calibrator.

simonpcouch:
Closes #893.

Affects:

* `tune_log()`
* `log_problems()`
* `.catch_and_log*()`

The `.catch_and_log*()` functions are exported, so will need revdepchecks.

### DIFF
--- a/R/checks.R
+++ b/R/checks.R
@@ -175,7 +175,7 @@ check_parameters <- function(wflow, pset = NULL, data, grid_names = character(0)
       msg <- paste0(msg, "s: ", paste0("'", unk_names, "'", collapse = ", "))
     }
 
-    tune_log(list(verbose = TRUE), split = NULL, msg, type = "info")
+    tune_log(list(verbose = TRUE), split_labels = NULL, msg, type = "info")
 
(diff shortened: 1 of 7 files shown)


simonpcouch:
agua is the only other package that uses `.catch_and_log()`:

https://github.com/tidymodels/agua/blob/6a742f637f3ae9ce70a7da117b8c6097355a0abf/R/tune.R#L105-L111

PR #919, opened 2024-07-18
use grid_space_filling() instead of grid_latin_hypercube()

topepo:
Depends on dev dials

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 
 * The package will now log a backtrace for errors and warnings that occur during tuning. When a tuning process encounters issues, see the new `trace` column in the `collect_notes(.Last.tune.result)` output to find precisely where the error occurred (#873).
 
-* When automatic grids are used, [dials::grid_space_filling()] is now used (instead of [dials::grid_latin_hypercube()]).
+* When automatic grids are used, `dials::grid_space_filling()` is now used (instead of `dials::grid_latin_hypercube()`).
 
 # tune 1.2.1
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
 
 * The package will now log a backtrace for errors and warnings that occur during tuning. When a tuning process encounters issues, see the new `trace` column in the `collect_notes(.Last.tune.result)` output to find precisely where the error occurred (#873).
 
+* When automatic grids are used, [dials::grid_space_filling()] is now used (instead of [dials::grid_latin_hypercube()]).
+
 # tune 1.2.1
 
 * Addressed issue in `int_pctl()` where the function would error when parallelized using `makePSOCKcluster()` (#885).
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 
 * The package will now log a backtrace for errors and warnings that occur during tuning. When a tuning process encounters issues, see the new `trace` column in the `collect_notes(.Last.tune.result)` output to find precisely where the error occurred (#873).
 
-* When automatic grids are used, `dials::grid_space_filling()` is now used (instead of `dials::grid_latin_hypercube()`).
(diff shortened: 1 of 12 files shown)


R/tune_bayes.R
```diff
@@ -632,7 +632,7 @@ fit_gp <- function(dat, pset, metric, eval_time = NULL, control, ...) {
 
 pred_gp <- function(object, pset, size = 5000, current = NULL, control) {
   pred_grid <-
-    dials::grid_latin_hypercube(pset, size = size) %>%
+    dials::grid_space_filling(pset, size = size, type = "latin_hypercube") %>%
```
topepo, line 635: I stuck with a LHD here because the new designs take a really long time to compute a really large grid (5,000 points).

NEWS.md
```diff
@@ -4,6 +4,8 @@
 
 * The package will now log a backtrace for errors and warnings that occur during tuning. When a tuning process encounters issues, see the new `trace` column in the `collect_notes(.Last.tune.result)` output to find precisely where the error occurred (#873).
 
+* When automatic grids are used, [dials::grid_space_filling()] is now used (instead of [dials::grid_latin_hypercube()]).
```
topepo, line 7: Do we know if this markup work in news files?

tests/testthat/_snaps/grid.md
```diff
@@ -66,14 +66,14 @@
...
+       1 <split [28/4]> Fold01 <tibble [6 x 5]> <tibble [0 x 4]>
+       2 <split [28/4]> Fold02 <tibble [6 x 5]> <tibble [0 x 4]>
+       3 <split [29/3]> Fold03 <tibble [6 x 5]> <tibble [0 x 4]>
+       4 <split [29/3]> Fold04 <tibble [6 x 5]> <tibble [0 x 4]>
+       5 <split [29/3]> Fold05 <tibble [6 x 5]> <tibble [0 x 4]>
+       6 <split [29/3]> Fold06 <tibble [6 x 5]> <tibble [0 x 4]>
+       7 <split [29/3]> Fold07 <tibble [6 x 5]> <tibble [0 x 4]>
+       8 <split [29/3]> Fold08 <tibble [6 x 5]> <tibble [0 x 4]>
+       9 <split [29/3]> Fold09 <tibble [6 x 5]> <tibble [0 x 4]>
+      10 <split [29/3]> Fold10 <tibble [6 x 5]> <tibble [0 x 4]>
```
simonpcouch, line 78: I'm surprised that this snap is changing. What's the explanation here?

NEWS.md
simonpcouch: Downlit will automatically link to docs when using backticks for pkgdown changelogs—let's use backticks in favor of the Roxygen2 formatting.

simonpcouch, line 7: Could we add a sentence or two on why we're doing this and also note that we still use a latin hypercube to generate grids for the GP?

tests/testthat/test-autoplot.R
```diff
@@ -329,12 +329,21 @@ test_that("plot_perf_vs_iter with fairness metrics (#773)", {
 
 test_that("regular grid plot", {
   skip_if_not_installed("ggplot2", minimum_version = "3.5.0")
```
simonpcouch, line 331: Was this test previously wrongly described? i.e. we weren't actually testing against `regular_grid()` before these changes, were we?

simonpcouch (approved):
Groovy🏄

tests/testthat/_snaps/grid.md
topepo: The "6" is because we have three standard metrics (added brier) but also different due to different initial grids.

simonpcouch: I understand that we have 3 default classification metrics now, but why wouldn't that snapped have failed before this PR? What do the grids have to do with this change?

tests/testthat/test-autoplot.R
topepo: Kind of. A one-point design is the same for any one-parameter design with one design point. I changed since the pre-optimized designs don't work in this situation. I'll put an issue in for that. @hfrick

PR #928, opened 2024-08-02
TDD prep: re`document()` with new roxygen

simonpcouch:
Also to ensure that GHA are running fine.


simonpcouch:
Going to close in favor of reviving #919.

PR #949, opened 2024-10-17
updates for no-suggests checks

topepo:
Add a GH workflow that only directly installs "hard" dependencies, i.e. Depends, Imports, and LinkingTo dependencies. Notably, Suggests dependencies are never installed, with the exception of testthat, knitr, and rmarkdown. The cache is never used to avoid accidentally restoring a cache containing a suggested dependency.


R/conf_mat_resampled.R
```diff
@@ -15,21 +15,23 @@
 #' library(parsnip)
 #' library(rsample)
 #' library(dplyr)
+#' if (rlang::is_installed("modeldata")) {
```
simonpcouch, line 18: Could we just use `@examplesIf` to avoid all of this indentation?

PR #950, opened 2024-10-23
add rlang type checkers

topepo:
Minimal changes, avoiding `rlang::abort()`, cli updates, and call routing 😞


R/checks.R
```diff
@@ -537,6 +517,7 @@ val_class_or_null <- function(x, cls = "numeric", where = NULL) {
   }
   invisible(NULL)
 }
+# TODO remove this once finetune is updated
```
topepo, line 520: Note to myself and @simonpcouch to make similar changes to finetune so that we can remove these three checking functions

R/checks.R
```diff
@@ -493,26 +493,6 @@ get_objective_name <- function(x, metrics) {
   x
 }
 
-
-# ------------------------------------------------------------------------------
-# acq functions
-
-check_direction <- function(x) {
```
topepo, line 500: DELETE

simonpcouch: SO SATISFYING

R/checks.R
```diff
@@ -644,3 +625,11 @@ check_eval_time <- function(eval_time, metrics) {
   invisible(NULL)
 
 }
+
+check_time_limit_arg <- function(x, call = rlang::caller_env()) {
+  if (!inherits(x, c("logical", "numeric")) || length(x) > 1) {
```
simonpcouch, line 630:
```suggestion
  if (!inherits(x, c("logical", "numeric")) || length(x) != 1L) {
```

As this ought not to be length-0 either, right?

tests/testthat/_snaps/bayes.md
```diff
@@ -393,12 +393,12 @@
...
       x Fold1: preprocessor 1/1:
         Error in `step_spline_b()`:
-        Caused by error in `glue()`:
-        ! Expecting '}'
+        Caused by error in `spline_msg()`:
+        ! Error in if (df < 0)  : missing value where TRUE/FALSE needed
       x Fold2: preprocessor 1/1:
         Error in `step_spline_b()`:
-        Caused by error in `glue()`:
-        ! Expecting '}'
```
simonpcouch, line 401: I'm seeing this change reverted in the snaps on GHA🤔

topepo: The error we should get is there now; I fixed a bug in recipes that resolved the `glue()` one. Funny: we were grepping against the glue issue in the logging tests here. I converted it to a snapshot.

PR #951, opened 2024-10-23
update tests based on dev ggplot2

topepo:
The ggplot2 news file implies that the plot object structure has changed a bit when it comes to labels: 

> (internal) default labels are derived in ggplot_build() rather than in ggplot_add.Layer() (@teunbrand, #5894)
> An attempt is made to use a variable's label attribute as default label (@teunbrand, #4631)

We have tests that inspect the labels and, if we are using the defaults, they are now missing. 

This PR will probably fail until #950 is merged. 

This probably affects finetune too @simonpcouch


teunbrand:
Hi Max, thanks for keeping vigilant and up to date! If you want to continue to test the labels, we also implemented an accessor (https://github.com/tidyverse/ggplot2/pull/6078) for exactly this reason.

topepo:
> we also implemented an accessor (https://github.com/tidyverse/ggplot2/pull/6078) for exactly this reason.

That, good sir, is exceedingly helpful. 

Is there anything similar for mappings and/or facets

teunbrand:
For strip labels we have a new `get_strip_labels()` function (https://github.com/tidyverse/ggplot2/pull/5924).
I don't think we have something for mappings, but I wouldn't oppose implementing something to that extent if you file a feature request.

hfrick:
Max, did you want to make use of that accessor that Teun mentioned? Or not at this time?

topepo:
It's in dev ggplot2 so we'll wait #952

hfrick (approved):

PR #957, opened 2024-11-01
refactor `compute_grid_info()`

simonpcouch:
Refactors `compute_grid_info()` to set the stage for postprocessing. Instead of having one function per `has_preprocessor x has_model x has_postprocessor` combination, simplifies to one code path.

Related to #948, where I made sure to test thoroughly that these changes would have no effect on output. The column ordering does indeed change, but this object never makes it to users and we always subset it by name internally.

### DIFF
--- a/R/grid_helpers.R
+++ b/R/grid_helpers.R
@@ -307,8 +307,6 @@ compute_grid_info <- function(workflow, grid) {
     return(out)
   }
 
-  grid <- tibble::as_tibble(grid)
-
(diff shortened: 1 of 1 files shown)


topepo (approved):
These changes make a lot of sense; we over-engineered the code that generates the grid tasks early on. 

I've tested this in a few different scenarios today and don't see any issues. 

Also, I would not worry about the column order; I doubt that we are calling by order instead of name (and shame on us if we do 😄 )

PR #960, opened 2024-11-04
address `compute_grid_info()` bug for partially regular grids

simonpcouch:
Closes #959.

PR #961, opened 2024-11-13
Compute grid info dplyr

### LINKED ISSUE #959: bug in refactored `compute_grid_info()`
After merging #957, we saw the following failures on extratests over the weekend:

```
══ Failed ══════════════════════════════════════════════════════════════════════
── 1. Failure ('test-mars-tuning.R:63:3'): tuning for mars() -- submodels *and* 
mars_metrics$.config (`actual`) not equal to paste0("Preprocessor1_Model", 1:7) (`expected`).

    actual                 | expected                  
[1] "Preprocessor1_Model1" | "Preprocessor1_Model1" [1]
[2] "Preprocessor1_Model1" - "Preprocessor1_Model2" [2]
[3] "Preprocessor1_Model2" - "Preprocessor1_Model3" [3]
[4] "Preprocessor1_Model2" - "Preprocessor1_Model4" [4]
[5] "Preprocessor1_Model3" - "Preprocessor1_Model5" [5]
[6] "Preprocessor1_Model4" - "Preprocessor1_Model6" [6]
[7] "Preprocessor1_Model5" - "Preprocessor1_Model7" [7]

── 2. Failure ('test-mars-tuning.R:68:3'): tuning for mars() -- submodels *and* 
count(mars_preds, .row, .config)$n (`actual`) not equal to rep(1L, 5537) (`expected`).

       actual | expected       
   [4] 1      | 1        [1]   
   [5] 1      | 1        [2]   
   [6] 2      - 1        [3]   
   [7] 2      - 1        [4]   
```

Some additional trickiness here in grids that are partially regular and partially not so.

topepo:
Closes #959.

A rework with Simon of #960

The grids are sorted in the new version, so there were some changes to the expected results.

### DIFF
--- a/R/grid_helpers.R
+++ b/R/grid_helpers.R
@@ -330,7 +330,7 @@ compute_grid_info <- function(workflow, grid) {
       dplyr::arrange(!!!syms_pre) %>%
       dplyr::mutate(
         .iter_preprocessor = dplyr::row_number(),
-        .lab_pre = recipes::names0(max(dplyr::n()), "Preprocessor")
+        .lab_pre = paste0("Preprocessor", 1:dplyr::n())
       )
     res <-
       dplyr::full_join(res, pp_df, by = parameters_preprocessor$id) %>%
@@ -377,7 +377,7 @@ make_iter_config <- function(dat) {
   # Compute labels for the models *within* each preprocessing loop.
   num_submodels <- purrr::map_int(dat$.submodels, ~ length(unlist(.x)))
   num_models <- sum(num_submodels + 1) # +1 for the model being trained
-  .mod_label <- recipes::names0(num_models, "Model")
+  .mod_label <- paste0("Model", 1:num_models)
   .iter_config <- paste(dat$.lab_pre[1], .mod_label, sep = "_")
   .iter_config <- vctrs::vec_chop(.iter_config, sizes = num_submodels + 1)
   tibble::tibble(.iter_config  = .iter_config)
--- a/R/grid_helpers.R
+++ b/R/grid_helpers.R
@@ -318,50 +318,69 @@ compute_grid_info <- function(workflow, grid) {
 
   res <- min_grid(extract_spec_parsnip(workflow), grid)
 
+  syms_pre <- rlang::syms(parameters_preprocessor$id)
+  syms_mod <- rlang::syms(parameters_model$id)
+
+  # ----------------------------------------------------------------------------
+  # Create an order of execution to train the preprocessor (if any). This will
+  # define a loop over any preprocessing tuning parameter combinations.
   if (any_parameters_preprocessor) {
-    res$.iter_preprocessor <- seq_len(nrow(res))
+    pp_df <-
+      dplyr::distinct(res, !!!syms_pre) %>%
+      dplyr::arrange(!!!syms_pre) %>%
+      dplyr::mutate(
+        .iter_preprocessor = dplyr::row_number(),
+        .lab_pre = recipes::names0(max(dplyr::n()), "Preprocessor")
+      )
+    res <-
+      dplyr::full_join(res, pp_df, by = parameters_preprocessor$id) %>%
+      dplyr::arrange(.iter_preprocessor)
   } else {
     res$.iter_preprocessor <- 1L
+    res$.lab_pre <- "Preprocessor1"
   }
 
+  # Make the label shown in the grid and in loggining
   res$.msg_preprocessor <-
     new_msgs_preprocessor(
-      seq_len(max(res$.iter_preprocessor)),
+      res$.iter_preprocessor,
       max(res$.iter_preprocessor)
     )
 
-  if (nrow(res) != nrow(grid) ||
-      (any_parameters_model && !any_parameters_preprocessor)) {
-    res$.iter_model <- seq_len(dplyr::n_distinct(res[parameters_model$id]))
-  } else {
-    res$.iter_model <- 1L
-  }
+  # ----------------------------------------------------------------------------
+  # Now make a similar iterator across models. Conditioning on each unique
+  # preprocessing candidate set, make an iterator for the model candidate sets
+  # (if any)
 
-  res$.iter_config <- list(list())
-  for (row in seq_len(nrow(res))) {
-    res$.iter_config[row] <- list(iter_config(res[row, ]))
-  }
+  res <-
+    res %>%
+    dplyr::group_nest(.iter_preprocessor, keep = TRUE) %>%
+    dplyr::mutate(
+      .iter_config = purrr::map(data, make_iter_config),
+      .model = purrr::map(data, ~ tibble::tibble(.iter_model = seq_len(nrow(.x))))
+    ) %>%
+    dplyr::select(-.iter_preprocessor) %>%
+    tidyr::unnest(cols = c(data, .model, .iter_config)) %>%
+    dplyr::select(-.lab_pre) %>%
+    dplyr::relocate(dplyr::starts_with(".iter"))
 
   res$.msg_model <-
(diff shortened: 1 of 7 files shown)


R/grid_helpers.R
```diff
@@ -318,50 +318,71 @@ compute_grid_info <- function(workflow, grid) {
...
+  # preprocessing candidate set, make an iterator for the model candidate sets
+  # (if any)
+
+  res <-
+    res %>%
+    dplyr::group_nest(.iter_preprocessor, keep = TRUE) %>%
+    dplyr::mutate(
+      .iter_config = purrr::map(data, make_iter_config),
+      .model = purrr::map(data, ~ tibble::tibble(.iter_model = seq_len(nrow(.x)))),
+      .num_models = purrr::map_int(.model, nrow)
```
topepo, line 361: This is needed for `res$.msg_model` (and is removed later)

R/grid_helpers.R
```diff
@@ -318,50 +318,71 @@ compute_grid_info <- function(workflow, grid) {
...
+    ) %>%
+    dplyr::select(-.iter_preprocessor) %>%
+    tidyr::unnest(cols = c(data, .model, .iter_config)) %>%
+    dplyr::select(-.lab_pre) %>%
+    dplyr::relocate(dplyr::starts_with(".iter"))
```
topepo, line 366: This (and another `relocate` below) should help with column ordering stability ~in case~ we refactor again.

R/grid_helpers.R
```diff
@@ -318,50 +318,71 @@ compute_grid_info <- function(workflow, grid) {
...
+  syms_mod <- rlang::syms(parameters_model$id)
+
+  # ----------------------------------------------------------------------------
+  # Create an order of execution to train the preprocessor (if any). This will
+  # define a loop over any preprocessing tuning parameter combinations.
   if (any_parameters_preprocessor) {
-    res$.iter_preprocessor <- seq_len(nrow(res))
+    pp_df <-
+      dplyr::distinct(res, !!!syms_pre) %>%
+      dplyr::arrange(!!!syms_pre) %>%
```
topepo, line 330: This will make the ordering of the preprocessors predictable. Previously, it would order them as-is. It's no big deal, but someone might wonder why `deg_free = 10` is executed before `deg_free=2`. 

This is why there are several `sort()` calls in the unit tests.

tests/testthat/test-grid_helpers.R
```diff
@@ -212,6 +221,12 @@ test_that("compute_grid_info - recipe and model (with and without submodels)", {
       list(trees = c(1L, 1000L))
     )
   )
+  expect_equal(
```
topepo, line 224: We can't be too paranoid about this.

tests/testthat/test-grid_helpers.R
```diff
@@ -185,25 +189,30 @@ test_that("compute_grid_info - recipe and model (with and without submodels)", {
   # use grid_regular to (partially) trigger submodel trick
   set.seed(1)
   param_set <- extract_parameter_set_dials(wflow)
-  grid <- bind_rows(grid_regular(param_set), grid_space_filling(param_set))
+  grid <-
```
topepo, line 192: I was going to add more tests for unbalanced grids but this one covers it well.

tests/testthat/test-grid_helpers.R
```diff
@@ -185,25 +189,30 @@ test_that("compute_grid_info - recipe and model (with and without submodels)", {
...
     unique(res$.msg_preprocessor),
     paste0("preprocessor ", 1:5, "/5")
   )
-  expect_equal(res$trees, c(rep(max(grid$trees), 10), 1))
+  expect_equal(sort(res$trees), sort(c(rep(max(grid$trees), 10), 1)))
   expect_equal(unique(res$.iter_model), 1:3)
   expect_equal(
-    res$.iter_config[1:3],
+    res$.iter_config[res$.iter_preprocessor == 1],
     list(
```
simonpcouch, line 206: Your new enumeration is correct (and fixes the case where I had broken previously), but it does look like we're missing a `format_with_padding` still. When I run these tests with CRAN `compute_grid_info()`, I see:

```
── Failure (test-grid_helpers.R:204:3): compute_grid_info - recipe and model (with and without submodels) ──
res$.iter_config[res$.iter_preprocessor == 1] (`actual`) not equal to list(...) (`expected`).

    actual[[1]]             | expected[[1]]             
[1] "Preprocessor1_Model01" - "Preprocessor1_Model1" [1]
[2] "Preprocessor1_Model02" - "Preprocessor1_Model2" [2]
[3] "Preprocessor1_Model03" - "Preprocessor1_Model3" [3]
[4] "Preprocessor1_Model04" - "Preprocessor1_Model4" [4]

`actual[[2]]`:   "Preprocessor1_Model05" "Preprocessor1_Model06" "Preprocessor1_Model07"
`expected[[2]]`: "Preprocessor1_Model5"  "Preprocessor1_Model6"  "Preprocessor1_Model7" 

`actual[[3]]`:   "Preprocessor1_Model08" "Preprocessor1_Model09" "Preprocessor1_Model10"
`expected[[3]]`: "Preprocessor1_Model8"  "Preprocessor1_Model9"  "Preprocessor1_Model10"
```

R/grid_helpers.R
```diff
@@ -318,50 +318,71 @@ compute_grid_info <- function(workflow, grid) {
...
-    "Preprocessor",
-    res_row$.iter_preprocessor,
-    "_Model",
-    format_with_padding(models)
-  )
+make_iter_config <- function(dat) {
+  # Compute labels for the models *within* each preprocessing loop.
+  num_submodels <- purrr::map_int(dat$.submodels, ~ length(unlist(.x)))
+  num_models <- sum(num_submodels + 1) # +1 for the model being trained
+  .mod_label <- paste0("Model", 1:num_models)
```
simonpcouch, line 382:
```suggestion
  .mod_label <- paste0("Model", seq_len(num_models))
```

simonpcouch: Would be curious if we could just add a `format_with_padding()` here, but I think that actually needs to happen outside of the loop since a different preprocessor might be paired with models that enum up to the tens / hundreds place.

simonpcouch (commented):
Awesome. This fixes bugs I had introduced in the previous refactor and I find it more readable. 

We're just one `format_with_padding()` away from being there, though it's proven tricky to situate that in the right place given that each of those `.iter_config`s live inside of lists. I'm open to entertaining the argument that the padding is ultimately unneeded if you think that's the best way to move forward—those `.iter_config` indices can be reasonably separated using just "Preprocessor" and "_Model" already. In the case we're fine with that padding going away, consider this an "Approve" merge once we've noted the change in NEWS. Otherwise, a bit more work to do.

I just adjusted the ref in https://github.com/tidymodels/extratests/pull/227 to test this PR. I'm assuming it will fail with a `format_with_padding()` snap change but otherwise run fine.

R/grid_helpers.R
topepo: IT's pretty easy with `recipes::names0()`: https://github.com/tidymodels/tune/commit/2a78da3477efc77a4fe64e128fcf7318b4c8dfb2#diff-f8b62de58f2adfdb08f139a17fd26e3ddb0b7a0ab33909f5bbdb6fdb93ba7923L333

topepo: I think that this will work in `make_iter_config()`; all the models within the preprocessor combination are present at that time.

topepo: tidymodels/extratests/227 looks goo after the last commit.

simonpcouch (approved):
Nice, I see passing tests with the CRAN version, too! Let's merge this in.

PR #965, opened 2024-11-14
Use a random grid for a one point design

### LINKED ISSUE #962: use an initial random design with grid size is 1
`tune:::create_initial_set()` will use `dials::grid_space_filling()` when an integer is passed. 

We have space-filling designs for 2+ points and when `grid = 1`, it creates a random grid btu uses a LHD and issues a warning: 

> Due to the small size of the grid, a Latin hypercube design will be used. 

We can get a random point via `grid_random()` without the warning. 

`finetune::tune_sim_anneal()` defaults to a 1-point initialization so no warning would be good.

topepo:
Closes #962

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,7 +4,7 @@
 
 * The package will now log a backtrace for errors and warnings that occur during tuning. When a tuning process encounters issues, see the new `trace` column in the `collect_notes(.Last.tune.result)` output to find precisely where the error occurred (#873).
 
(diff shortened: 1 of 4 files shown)


hfrick:
We don't need this PR anymore, given that `dials::grid_space_filling()` no longer warns about `n = 1`, do we? 

dials PR with the relevant change: https://github.com/tidymodels/dials/pull/363

PR #966, opened 2024-11-15
enable tuning postprocessors

simonpcouch:
**Heads up--this PR is not yet functional!**

Enables tuning arguments in tailor postprocessors. Works in some cases, but still fails most others. Importantly, this disables the submodel trick right now and will need to be addressed before this is ready for review. [Here are some notes](https://gist.github.com/simonpcouch/28d984cdcc3fc6d22ff776ed8740004e) about what's so tricky there.

### DIFF
--- a/R/grid_helpers.R
+++ b/R/grid_helpers.R
@@ -1,7 +1,6 @@
 
 predict_model <- function(new_data, orig_rows, workflow, grid, metrics,
                           submodels = NULL, metrics_info, eval_time = NULL) {
-
   model <- extract_fit_parsnip(workflow)
 
   forged <- forge_from_workflow(new_data, workflow)
@@ -260,6 +259,22 @@ finalize_workflow_preprocessor <- function(workflow, grid_preprocessor) {
   workflow
 }
 
+#' @export
+#' @rdname tune-internal-functions
+finalize_workflow_postprocessor <- function(workflow, grid_postprocessor) {
+  # Already finalized, nothing to tune
+  if (ncol(grid_postprocessor) == 0L) {
+    return(workflow)
+  }
+
+  postprocessor <- workflows::extract_postprocessor(workflow)
+  postprocessor <- merge(postprocessor, grid_postprocessor)$x[[1]]
+
+  workflow <- set_workflow_tailor(workflow, postprocessor)
+
+  workflow
+}
+
 # ------------------------------------------------------------------------------
 
 # For any type of tuning, and for fit-resamples, we generate a unified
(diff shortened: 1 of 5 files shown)


R/grid_helpers.R
```diff
@@ -351,6 +370,17 @@ compute_grid_info <- function(workflow, grid) {
   # Now make a similar iterator across models. Conditioning on each unique
   # preprocessing candidate set, make an iterator for the model candidate sets
   # (if any)
+  if (any_parameters_postprocessor) {
+    # Ensure that the submodel trick kicks in by temporarily nesting the
+    # postprocessor parameters while iterating in the model grid
+    # TODO: will this introduce issues when there are matching postprocessor
+    # values across models?
+    # ... i think we actually want to (temporarily?) situate these as submodels
```
simonpcouch, line 378: My first pass at enabling the submodel trick—nest by the postprocessor and unnest later—but that doesn't quite do the job.

R/min_grid.R
```diff
@@ -326,3 +326,13 @@ min_grid.pls <- fit_max_value
...
 min_grid.poisson_reg <- fit_max_value
+
+
+# When `min_grid()` is applied to grids with additional columns for
+# postprocessors, we need to nest the postprocessor columns into
+# .submodels to effectively enable the submodel trick.
+# See: https://gist.github.com/simonpcouch/28d984cdcc3fc6d22ff776ed8740004e
+nest_min_grid <- function(min_grid, post_params) {
+  # TODO
+  min_grid
```
simonpcouch, line 337: I thiiink this will be our entry point; patch `min_grid()` output by dropping postprocessor values into the nested `.submodel` structure.

R/grid_helpers.R
```diff
@@ -385,6 +434,32 @@ make_iter_config <- function(dat) {
...
+      )
+    ) %>%
+    select(-.iter_config)
+}
+
+make_iter_config_post <- function(iter_config, iter_postprocessor) {
+  paste0(
+    iter_config,
+    "_Postprocessor",
+    iter_postprocessor
```
simonpcouch, line 459: Still needs a `format_with_padding()`.

R/grid_helpers.R
topepo: > that doesn't quite do the job.

Can you expand on this?

R/min_grid.R
topepo: The postprocessing object will only be available from within a workflow.  

We could make a workflow method for `min_grid()` so that we have all of the information coming from a single object. 

It might be possible to execute `min_grid()` on the model spec then generate a separate (but similar) data structure from the postprocessor, then join the two.

R/grid_helpers.R
simonpcouch: See the linked notes above!

simonpcouch:
Spent a bit with 4c763ed in an attempt to enable the submodel trick but still not quite there. None of the tuning code now works from this PR as I was focused on `compute_grid_info()` but need to put this away and focus elsewhere.

R/min_grid.R
simonpcouch: I wonder if a join, like you suggest, would be the more effective way to make this happen. Right now, my theory is that we can take the output from `model_spec` `min_grid()` methods and do some dplyr/tidyr-fu based on the postprocessor parameter names.

This is gnaaaarly, though.

R/grid_helpers.R
```diff
@@ -370,21 +391,74 @@ compute_grid_info <- function(workflow, grid) {
...
+        msgs_model = .msg_model
+      ),
+      .iter_config_post = purrr::map2(
+        .iter_config,
+        .iter_postprocessor,
+        make_iter_config_post
+      )
+    ) %>%
+    select(-.iter_config) %>%
+    nest(post = c(any_of(post_params), ".iter_postprocessor", ".msg_postprocessor", ".iter_config_post"))
```
simonpcouch, line 451: We actually want postprocessor parameters names both inside and outside of `.submodels`, but this results in them only being inside of `.submodels`.

R/grid_helpers.R
topepo: Those were very helpful. Thanks

Issue #660, opened 2023-04-04
`tune_grid()` fails if parameter is not marked as `tunable()` despite matching grid


hfrick:
In #633, we enabled tune to deal with list-columns in the grid, in order to allow people to use a custom grid to tune lower-level arguments.

The example there was based on wanting to tune parameters to a custom loss function that was passed to the `objective` engine argument for xgboost. When applying the same approach to a recipe step, it fails:

``` r
library(tidymodels)
data("credit_data", package = "modeldata")

set.seed(342)
credit_resamples <- vfold_cv(credit_data, v = 5)

rec <- recipe(Price ~ ., data = credit_data) %>%
  step_impute_bag(
    Status, Home, Marital, Job, Income, Assets, Debt,
    # instead of 
    # options = list(nbagg = tune())
    # do
    options = tune()
  )

credit_wflow <- workflow(rec, linear_reg())

# make grid manually
grid_options <- tibble(nbagg = seq(15, 30, by = 5)) %>% 
  mutate(options = map(nbagg, ~ list(nbagg = .))) %>% 
  select(options)

credit_res <- tune_grid(credit_wflow, credit_resamples, grid = grid_options)
#> Warning: No tuning parameters have been detected, performance will be evaluated
#> using the resamples with no tuning. Did you want to [tune()] parameters?
#> → A | error:   You cannot `prep()` a tuneable recipe. Argument(s) with `tune()`: 'options'. Do you want to use a tuning function such as `tune_grid()`?
#> There were issues with some computations   A: x1
#> There were issues with some computations   A: x5
#> 
#> Warning: All models failed. Run `show_notes(.Last.tune.result)` for more
#> information.
```

<sup>Created on 2023-04-04 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

More details on why this fails are below but the higher-level question is: If you tag something for tuning and bring your own grid for it, shouldn't it work? 

Currently, this traces back to `tune_args()` for recipes limiting the scope of checking for `tune()` tags to arguments which were designated for tuning via `tunable()`.

https://github.com/tidymodels/recipes/blob/567fe17cddb2cb880bcd30daa7737effca065ed9/R/tune_args.R#L29-L32

We can bypass some previous checks by giving `tune_grid()` a decoy parameter set but because `tune_args()`, called here:
https://github.com/tidymodels/tune/blob/802909c7aa4c554b3890b7ff23a25bfd80e601dd/R/checks.R#L75

[...]

PR #974, opened 2024-12-18
Tune schedule

topepo:
This draft PR is for testing a new data structure that governs how tune will efficiently process grids. Two new functions are currently exported for debugging but will not be exported in the final version: 

 - `get_tune_schedule()` produces a tibble structure with nested levels that contains information about what tuning parameters should be used at different stages of execution. This PR is focused on making sure that this structure accounts for all of the tuning scenarios. 
 - `loopy()` is a temporary function that emulates `tune_grid_loop_iter()` (without the bells and whistles) that is used to test `get_tune_schedule()` and also to understand how we will incorporate postprocessors into the mix.


hfrick:
My high-level thoughts on `get_schedule()`
- Introducing `pre_stage`
	- We do not need another nesting structure since the preprocessing stage is the "top-level" stage in this context but I'm wondering if it would be a good idea nonetheless: Currently, all the other stages have a named (nested) tibble associated with them, even if the tibble is empty (i.e., has zero rows). That makes it easier to understand the object structure imo which is my main motivation for bringing this up. Bonus: it makes all the stages feel the same / have a corresponding structure.
- Breaking up `get_schedule()` 
	- Into `schedule_proproc()`, `schedule_model_stage()`, `schedule_predict_stage()` which would all schedule "their" stage and push the rest of the candidate values into a tibble for the next stage. We'd call them recursively.
	- This should make it easier to test the scheduling components separately and then their nesting within each other.
	- It would also make it easier to add  a `schedule_postproc()` if we ever want to get more complex about how to schedule in this stage. (I'm thinking about whether or not we allow more flexibility in the order of post-processing adjustments and/or add more adjustments.) Or tinker with how to schedule the preprocessing in a future with more info on the order of recipe steps?
- Keeping the scheduling direction linear
	- This is the thing that struck me already when reading the corresponding writeup. 
	- I'd prefer to start "outside" with preprocessing and end with post-stage, not go from model-stage to post-stage and then do the pre-stage.
	- Part of that would be to use `min_grid()` only on the model parameters, without preprocessing or post-processing parameters, essentially using it as, or creating a, `minimal_model_grid()`.

topepo:
Note: let's take a look at how this structure can work with agua, which does the model fits _en masse_ via h2o.

hfrick:
Another note: We currently use the parameter set to determine which parameter, i.e., column in the grid, is associated with the preprocessing, model, or post-processing stage (via the `$source` column). With https://github.com/tidymodels/tune/issues/660 in mind, we might want to try to get to that information without adding another dependency on the parameter set. Using `tune_args(wflow)` might be a solution, although, ideally, we want something that tells us the location of the `tune()` placeholders in the workflow, and judging from #660 that's not always what `tune_args()` does either. (But maybe should? - That's a different issue though.)

topepo:
For agua... the current routine for processing the grid normally is [here](https://github.com/tidymodels/tune/blob/26d6e213b10e6639ca710d1698fcb2f8b2d9574f/R/grid_code_paths.R#L337) and the agua equivalent is [here](https://github.com/tidymodels/agua/blob/dfc7a13ad37d5ddbe58cb4d37f6b4de494118a20/R/tune.R#L30). Basically, agua consolidates the work that `tune_grid()` does via loop into a single function call. 

For `loopy()`, I think that we would have to take a similar strategy. With `loopy()`, there isn't a top-level postprocessing loop within the model loop. That's because of the three cases mentioned [here](https://github.com/tidymodels/tune/blob/26d6e213b10e6639ca710d1698fcb2f8b2d9574f/R/loopy.R#L118). For agua, we are likely to need a post-processing loop since we have all of the model results at once, and the post-actions will happen on a model-by-model basis. 

TL;DR

agua will work similarly to tune, except instead of losing a loop, the loss of the model loop is offset by a post loop. It should still be very efficient.

hfrick:
@topepo Also okay to close this one, right?

PR #978, opened 2025-01-17
Refactor of `get_tune_schedule()`

hfrick:
Here's the refactor of `get_tune_schedule()`! The basic idea is to schedule the stages recursively, starting at the preprocessing stage down to the postprocessing stage, and always do one stage at a time, pushing the remaining parameters into a nested tibble.

I've made a PR into `tune-schedule` so that you can see the diffs to the previous version clearly. I understand that branch to be our place to work things out, so I'm happy to make a separate PR into `main` if we are happy with how `get_tune_schedule()` looks. I think the tests could also make use of that separation of the stages into smaller scheduling functions, but I didn't do this here because I wanted you to see how the tests changed for this refactor.

The refactor leads to 0-row tibbles when there are no tuning parameters at all (which we discussed in the team meeting) and small changes in the order of the columns. The ordering of the rows (for preprocessing) also stays the same now between the ingoing grid and the outgoing schedule.

Since this is the second round of working over this scheduling function, no need to review "only" high-level, hit me with your nits so that this part is ready for main!


R/schedule.R
```diff
@@ -21,132 +21,109 @@ get_tune_schedule <- function(wflow, param, grid) {
...
+    pull(id)
 
-	sched <- sched %>% dplyr::group_nest(!!!symbs$pre, .key = "model_stage")
+  predict_stage %>% 
+    tidyr::nest(.by = all_of(submodel_param), .key = "post_stage")
+}
 
-	# ------------------------------------------------------------------------------
+get_param_info <- function(wflow) {
+  param_info <- tune_args(wflow) %>%
```
hfrick, line 115: Using `tune_args()` here instead of a parameter set object, due to considerations I've put in https://github.com/tidymodels/tune/pull/974#issuecomment-2590393148

tests/testthat/helper-tune-package.R
```diff
@@ -83,33 +87,33 @@ redefer_initialize_catalog <- function(test_env) {
...
 
 
-mod_tune_bst <- boost_tree(trees = tune(), min_n = tune(), mode = "regression")
-mod_tune_rf <- rand_forest(min_n = tune(), mode = "regression")
+mod_tune_bst <- parsnip::boost_tree(trees = tune(), min_n = tune(), mode = "regression")
+mod_tune_rf <- parsnip::rand_forest(min_n = tune(), mode = "regression")
 
 if (rlang::is_installed("probably")) {
 
 	adjust_tune_min <-
```
hfrick, line 107: Given that we usually use `rec` in the name of recipes objects, I would like to advocate for calling tailor objects something with `tailor` rather than `adjust_`.

simonpcouch (commented):
Going to hold off on a proper review until I can carve out a solid chunk of time, but re:

> I've made a PR into tune-schedule so that you can see the diffs to the previous version clearly. I understand that branch to be our place to work things out, so I'm happy to make a separate PR into main if we are happy with how get_tune_schedule() looks. I think the tests could also make use of that separation of the stages into smaller scheduling functions, but I didn't do this here because I wanted you to see how the tests changed for this refactor.

Totally makes sense, thanks! I'm definitely on board for the workflow of taking chunks of that PR and refactoring + reviewing more in-detail and then sending those smaller portions into `main` as we do so.

Just eyeballing the diffs, it looks like this PR still makes use of the `UseSpacesForTab: No` setting. I see that it probably makes sense to keep that setting around to prevent conflicts with—and more easily diff against—`tune-schedule`, but I'd advocate for reverting back to `UseSpacesForTab: Yes` and reformatting the smaller chunks at some point before we send them into `main`. I can imagine a couple different ways that workflow could look (wait to reformat, merge into `tune-schedule`, reformat that whole PR to line up with the rest of the repo, extract out the relevant bits and merge to `main`?), but whatever results in the least work for the implementer has a thumbs-up from me.

simonpcouch (approved):
Very readable, very concise. A huge step up from `compute_grid_info()` or any of its refactors. Got a lot of joy from reviewing this one—bravo to yall!

+1 to working in some tests at the level of the newly separate functions, but fine with me to wait for a separate PR to make that happen.

Huzzah🙆

tests/testthat/test-schedule.R
```diff
@@ -3,6 +3,14 @@
...
 # Objects in helper-tune-package.R
 
+suppressPackageStartupMessages(library(workflows))
+suppressPackageStartupMessages(library(parsnip))
+suppressPackageStartupMessages(library(recipes))
+suppressPackageStartupMessages(library(dials))
+suppressPackageStartupMessages(library(tailor))
+suppressPackageStartupMessages(library(purrr))
+suppressPackageStartupMessages(library(dplyr))
+
```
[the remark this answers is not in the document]
hfrick: Repeating the library calls in each `test_that()` expression is actually my least favorite approach to this problem 😅  

Because: 
At least when executing such test code in the active console, the library calls have a lasting effect - but are less visible for quick cross-file scanning when they (unintentionally) affect something elsewhere. It could be that testthat sandboxes this behaviour but I think it might not.

``` r
library(testthat) 

test_that("tune function is not found", {
  expect_error(
    tune_grid(),
    "could not find"
  )
})
#> Test passed 🎊

test_that("find it with library", {
  library(tune)
  expect_error(
    tune_grid(),
    "The first argument"
  )
})
#> Test passed 🌈

test_that("tune function is still not found", {
  expect_error(
    tune_grid(),
    "could not find"
  )
})
#> ── Failure: tune function is still not found ───────────────────────────────────
#> `tune_grid()` threw an error with unexpected message.
#> Expected match: "could not find"
#> Actual message: "The first argument to [tune_grid()] should be either a model or workflow."
#> Backtrace:
#>     ▆
#>  1. ├─testthat::expect_error(tune_grid(), "could not find")
#>  2. │ └─testthat:::quasi_capture(...)
#>  3. │   ├─testthat (local) .capture(...)
#>  4. │   │ └─base::withCallingHandlers(...)
#>  5. │   └─rlang::eval_bare(quo_get_expr(.quo), quo_get_env(.quo))
#>  6. ├─tune::tune_grid()
#>  7. └─tune:::tune_grid.default()
#>  8.   └─rlang::abort(msg)
#> Error:
#> ! Test failed
```

<sup>Created on 2025-01-21 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

R/schedule.R
```diff
@@ -21,132 +21,109 @@ get_tune_schedule <- function(wflow, param, grid) {
...
-		)
+schedule_stages <- function(grid, param_info, wflow) {
+	# schedule preprocessing stage and push the rest into a nested tibble
+	param_pre_stage <- param_info %>% 
+    filter(source == "recipe") %>% 
+    pull(id)
+	schedule <- grid %>% 
+    tidyr::nest(.by = all_of(param_pre_stage), .key = "model_stage")
+
+	# schedule next stages recursively
```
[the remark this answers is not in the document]
hfrick: A) > hit me with your nits 

B) Not a nit, but a valuable comment! You're right, I supposed it's not quite the right word. What would you call it? Something with "nested"? Or just "schedule next stages within `schedule_model_stage_i()"? I just want to give the reader a heads-up that all stages will be taken care of, even though you can only "see" scheduling the immediate next stage from that point in the code.

R/schedule.R
```diff
@@ -21,132 +21,109 @@ get_tune_schedule <- function(wflow, param, grid) {
...
+					schedule_model_stage_i, 
+					param_info = param_info,
+					wflow = wflow
+				)
+    )
+}
 
-	# ------------------------------------------------------------------------------
-	# Merge in submodel fit value (if any)
+schedule_model_stage_i <- function(model_stage, param_info, wflow){
```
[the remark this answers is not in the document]
hfrick: Max worked it out and now I got to let it shine 😄

R/schedule.R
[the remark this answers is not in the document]
simonpcouch: Yeah, I see what you're trying to call out! Maybe "nested iteration"? Or possibly just point out "each model stage will also iterate across predict and post stages"

R/schedule.R
```diff
@@ -21,132 +21,109 @@ get_tune_schedule <- function(wflow, param, grid) {
...
+}
 
-	sched <- sched %>%
-		dplyr::mutate(
-			predict_stage = purrr::map(
-				predict_stage,
-				~.x %>% dplyr::group_nest(!!!symbs$sub, .key = "post_stage")
-			)
-		)
+schedule_stages <- function(grid, param_info, wflow) {
```
topepo, line 45: Do we need `param_info` as an argument? Since it is created by `get_param_info()`, we could call that immediately with `wflow`.

R/schedule.R
```diff
@@ -21,132 +21,109 @@ get_tune_schedule <- function(wflow, param, grid) {
...
-		sub_id <- param$id[param$source == "model_spec" & param$has_submodel]
-		non_sub_id <- param$id[param$source == "model_spec" & !param$has_submodel]
-	} else {
-		model_id <- sub_id <- non_sub_id <- character(0)
-	}
-
-	if (any(param$source == "tailor")) {
-		post_id <- param$id[param$source == "tailor"]
+	og_cls <- class(schedule)
+	if (nrow(param) == 0) {
```
topepo, line 30: It's a little gross to code here but this class structure feels like a good solution (until we know that it is not)

topepo (commented):
Also, since I forgot to say it in the review... this looks great. Big improvement on may refactor.

R/schedule.R
```diff
@@ -21,132 +21,109 @@ get_tune_schedule <- function(wflow, param, grid) {
...
+  non_submodel_param <- param_info %>% 
+    filter(source == "model_spec" & !has_submodel) %>% 
+    pull(id)
+  
+  # schedule model parameters
+  schedule <- min_model_grid(model_stage, model_param, wflow)
+
+  # push remaining paramters into the next stage
+  next_stage <- model_stage %>% 
+    tidyr::nest(.by = all_of(non_submodel_param), .key = "predict_stage")
```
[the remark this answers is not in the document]
hfrick: Thanks for all the code suggestions! I find it totally okay for the reviewer to do one example and then hand it back to whoever opened the PR. 

About the worker process: Do we expect to execute `get_tune_schedule()` in parallel? My understanding is that we first schedule and _then_ send todos off to workers, i.e. we would not expect to call `get_tune_schedule()` in parallel, no?

R/schedule.R
topepo: I think we keep it, default it to NULL and, if not NULL, run `get_param_info()`. 

It occurred to me that users can pass `param_info` in and that might differ from what we get by running `get_param_info(wflow)`

hfrick: We do not need the parameter set `param` as an input - good catch, thank you! 

What we do need is the grid to schedule, the info which parameter belongs to which stage, and the info which parameters are submodel paramters. For that last one, we need to know the model type and we deduct that from the workflow. The info which parameter belongs to which stage we originally pulled from the paramter set, that's why it's (still) in the function signature here.

But we can get that info also from the workflow, via `tune_args()`, which is what I've done here. My reasons for that:
- If someone provides a workflow with `tune()` tags and a corresponding grid, there is no (conceptual) need for a dials parameter set. Hence, me not wanting to rely on one. 
- If someone provides a custom parameter set, they still need to tag parameters via `tune()` in the workflow spec for anything to be tuned. Therefore, we can rely on the workflow via `tune_args()` and don't need a parameter set.

Therefore, my suggestion is use only `grid` and `wflow` in the signature here.

hfrick: Ah, right. Let's create it here, once, and then pass it through the different stages. That makes it cleaner at this level. I don't think it's helpful to recreate it in all following scheduling stages becase, in my thinking, the (primary) information I need for scheduling a parameter grid is which parameter belongs into which stage (which is what `param_info` captures). The workflow we only need for the model stage and only gets carried through until then.

Regarding users passing in a parameter set object, I'll add some thoughts to the signature of `get_tune_schedule()`.

hfrick:
Can't comment directly on the line, so putting this here:

While thinking about `param_info` in the signature of `schedule_stages()`, I realized that we do not need the parameter set `param` here as an input.

What we do need is the grid to schedule, the info which parameter belongs to which stage, and the info which parameters are submodel paramters. For that last one, we need to know the model type and we deduct that from the workflow. The info which parameter belongs to which stage we originally pulled from the paramter set, that's why it's (still) in the function signature here.

But we can get that info also from the workflow, via `tune_args()`, which is what I've done here. My reasons for that:
- If someone provides a workflow with `tune()` tags and a corresponding grid, there is no (conceptual) need for a dials parameter set. Hence, me not wanting to rely on one. 
- If someone provides a custom parameter set, they still need to tag parameters via `tune()` in the workflow spec for anything to be tuned. Therefore, we can rely on the workflow via `tune_args()` and don't need a parameter set.

Therefore, my suggestion is use only `grid` and `wflow` in the signature here.

topepo:
Should this be closed since #978 was merged?

hfrick:
yes!

PR #983, opened 2025-02-20
v1.2.2 release candidate

simonpcouch:
🚧 **This PR is not intended to be merged.** 🚧

This PR branches off of the [v1.2.1 tag](https://github.com/tidymodels/tune/releases/tag/v1.2.1), cherry-picks https://github.com/tidymodels/tune/commit/73692f4c851be423ebb8c5f819c812dc2c3e7fda, fixes a merge conflict, and compares to a branch based off the of the current package released with `git checkout -b released-1.2.1 v1.2.1`. The idea is that we'd send off a release of the package that just advances this deprecation so that we can implement the schedule functionality without having to accommodate foreach wonk.

PR #985, opened 2025-02-21
updates for new CRAN versions (esp dials)

topepo:
Recent CRAN versions of our package have broken a few unit tests. I'm mostly the culprit due to tidymodels/dials#370. 

We use KNN for our pre-defined objects, which are used in testing. The grid is a bit different and that propagates through the tests. 

There are also some overdue call threading changes (presumably from hardhat and/or parsnip). 

I'll have to funnel these changes to #983.


hfrick (approved):
Looks good!

PR #994, opened 2025-03-20
Add outcome_names() method for workflows using add_variables()

### LINKED ISSUE #993: cannot get outcome names for workflows using add_variables()
``` r
library(tidymodels)

lm_mod <- parsnip::linear_reg() %>% parsnip::set_engine("lm")
wflow <- workflow() %>% add_model(lm_mod)

wflow_1 <- wflow %>% add_variables(outcomes = "mpg", predictors = c(wt))

outcome_names(wflow_1)
#> Error in UseMethod("outcome_names"): no applicable method for 'outcome_names' applied to an object of class "workflow_variables"
```

<sup>Created on 2025-03-20 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

topepo:
Closes #993 993


R/outcome-names.R
```diff
@@ -39,20 +41,39 @@ outcome_names.recipe <- function(x, ...) {
...
+			call = call
+		)
+	}
+	res <- rlang::eval_tidy(x$outcomes, data, env = call)
+	res
+}
+
+#' @export
+#' @rdname outcome_names
+outcome_names.tune_results <- function(x, ..., call = caller_env()) {
```
hfrick, line 76: Do we need this change in function signature for `tune_results` objects? 🤔

R/outcome-names.R
```diff
@@ -39,20 +41,39 @@ outcome_names.recipe <- function(x, ...) {
...
+	...,
+	call = caller_env()
+) {
+	if (is.null(data)) {
+		cli::cli_abort(
+			"To determine the outcome names when {.fn add_variables} is used, please
+       pass the training set data to the {.arg data} argument.",
+			call = call
+		)
+	}
```
hfrick, line 69: Is this an error message for the user or for us?

hfrick (commented):
I'm missing some of the context in which this is (going to be) called, so I got two questions :D

R/outcome-names.R
topepo: You're welcome to suggest a change.

hfrick, line 76:
```suggestion
outcome_names.tune_results <- function(x, ...) {
```

I don't think the addition is necessary (and tests pass without it) so I'd just leave things as is.

R/outcome-names.R
hfrick: Okay, digging around a bit more here and I think there is a little messaging mismatch to resolve in that error message:

Do you mean that the user of `outcome_names()` should use the `data` argument of `outcome_names()`? The user of `outcome_names()` is presumably just us since `outcome_names()` is labeled with `@keywords internal`. In that case, I wouldn't pass in the call to `cli_abort()` and possibly remove the reference to `add_variables()`. The context of how we got into this probably lives better elsewhere, likely wherever we call it from and need to add `data`, so in `tune_grid()` or `int_pctl()`.

If you mean the general tidymodels user the current error message is probably confusing: they only call `tune_grid()` or `int_pctl()` and couldn't use that `data` argument to `outcome_names()`. `tune_grid()` does not have a `data` argument and `int_pctl()` has `.data` but that is not the same as the `data` for `outcome_name()`. (And imagining a very desperate user, looking into the mentioned `add_variables()` also does not reveal a `data` argument.) 

I'm guessing path 1 is the right one here?

hfrick (approved):

PR #1007, opened 2025-05-23
Add Fold Weights for Variable Resample Weighting

### LINKED ISSUE #990: Variable Fold Weights
I'm currently working through a project where there's time dependency, so I'm using expanding-window folds (e.g. fold 1 is from 2018-2021 with test set 2022, fold 2 is 2018-2022 with test set 2023, etc.). When running a hyperparameter optimization, to my understanding the code will create an average value of the evaluation metric, weighing each of these folds equivalently, even though they have differing number of records. In this case, it would probably make more sense to weight their contribution to the evaluation metric proportional to the number of instances in the given fold.

An example might be:

Fold | N | Metric
-- | -- | --
1 | 100 | 0.80
2 | 200 | 0.85
3 | 250 | 0.90
4 | 300 | 0.95

We should weigh fold 4 more than fold 1 when deciding the hyperparameters. So, under equal weighting we would get 0.875, whereas under weighting proportional to N would give 0.894.

I'm happy to work on this myself and make a PR, just wanted to confirm it would be a supported feature before I invested time into it.

As far as implementation, I think the straightforward way would be to pass a vector to the various `tune` functions (`tune_bayes` or `tune_grid`) indicating weights per-fold, and possibly make a helper function that creates that vector for you based on fold population, which seems like the most obvious case where you would use this. I could also see some argument for putting it into the `rset` object, but that might be a bit over-engineered.

tjburch:
Closes #990 

This PR implements support for variable fold weights in hyperparameter tuning. This is useful in cases where folds may have differing numbers of observations, and you want proportional contribution to hyperparameter selection.

The implementation adds two main functions: `add_fold_weights()` to attach custom weights to rset objects, and `calculate_fold_weights()` to automatically compute weights proportional to fold sizes. Weights are stored as .fold_weights attributes and should flow through the existing tuning pipeline. 

Core changes are in estimate_tune_results() which now detects weights and uses weighted statistics (weighted mean, weighted standard deviation, effective sample size) when aggregating metrics. Implementation should be backwards compatible and non-breaking.


topepo:
Hey @tjburch. Thanks for the PR.

We're doing a pretty invasive update the this package that will take some time. We'll look at the PR after things are settled there but it might be another 2-3 weeks. 

Is this something time-sensitive for you?

tjburch:
Nope. Just had some bandwidth staying awake on paternity leave. Review at your leisure, let me know if I can assist otherwise.

topepo:
> Just had some bandwidth staying awake on paternity leave

I estimate that 5% of all my work has been while waiting at the bustop or for some sort of practice to end 😄

tjburch:
Got around to deploying this into a local project and running into some odd errors. Will circle back.

R/checks.R
```diff
@@ -22,6 +22,26 @@ check_rset <- function(x) {
...
+
+  invisible(NULL)
+}
+
+#' Check fold weights in rset objects
+#'
+#' @param x An rset object.
+#' @return `NULL` invisibly, or error if weights are invalid.
+#' @keywords internal
+check_fold_weights <- function(x) {
```
topepo: Should we check for _equal_ weights too and return NULL if that is the case? I think that will be the case in some situations, and it would slightly increase efficiency to avoid work whenever possible.

R/utils.R
```diff
@@ -247,6 +247,284 @@ pretty.tune_results <- function(x, ...) {
...
+  }
+
+  # Add weights to the unique ID combinations
+  unique_ids$.fold_weight <- weights
+  unique_ids
+}
+
+#' @export
+#' @rdname fold_weights_utils
+.weighted_mean <- function(x, w) {
```
topepo: Why not use `stats ::weighted.mean()`

tjburch:
```
── Error ('test-checks.R:673:3'): fold weights with tune_grid ──────────────────
Error in `tune_grid(tune_wflow, weighted_folds, grid = simple_grid, control = control_grid(save_pred = FALSE))`: Package install is required for glmnet.
```

Looks like I'll need to replace the model here to something the ci tools have

EmilHvitfeldt:
> Looks like I'll need to replace the model here to something the ci tools have

yes that is correct! {xgboost} is a engine you can use, it is used in other tests

R/utils.R
topepo: So looking further down, we do the same with the variance calculation. We do have an _unexported_ API in recipes [`wt_calcs()`](https://github.com/tidymodels/recipes/blob/main/R/case_weights.R#L124) that does a lot of this. 

We can copy it over here (to avoid duplication) and think about moving that core code to hardat so that we get it everywhere from one source.

tests/testthat/test-checks.R
```diff
@@ -541,3 +541,163 @@ test_that("check parameter finalization", {
...
+  )
+
+  expect_error(
+    add_fold_weights(folds, c(-0.1, 0.5, 0.6)),
+    "must be non-negative"
+  )
+})
+
+test_that("fold weights integration test", {
+  skip_if_not_installed("rsample")
```
topepo: These are all imported so no need for these skips

tests/testthat/test-checks.R
```diff
@@ -541,3 +541,163 @@ test_that("check parameter finalization", {
...
+
+  # Test fold size calculation
+  calculated_weights <- calculate_fold_weights(folds)
+  expect_length(calculated_weights, nrow(folds))
+  expect_true(all(calculated_weights > 0))
+  expect_equal(sum(calculated_weights), 1) # Should sum to 1 now
+})
+
+test_that("fold weights with tune_grid", {
+  skip_if_not_installed("rsample")
```
topepo: Same as ☝️  etc

R/utils.R
topepo: _Erm actually_, we should probably put those weighting functions in a ["standalone file"](https://usethis.r-lib.org/reference/use_standalone.html) here, and then recipes and anyone else can import that. 

@EmilHvitfeldt do you have any thoughts on that?

topepo:
The testing issue for: 

> namespace 'workflowsets' is not available and has been replaced

is a false alarm. That is fixed in #1104.

R/utils.R
EmilHvitfeldt: I would be happy to have a standalone file for weighted functions

topepo: I'll amend my comments _a third time_. We don't export `wt_calcs()` but we do export the single statistic versions like `recipes::averages()` and `recipes::variances()` so maybe it's that simple.

topepo:
@tjburch I think that I'd like to change "fold weights" to "resample weights" to be more general. Is the former common terminology for these?

R/utils.R
topepo: @tjburch I made a change to use the functions from the stats package. Is the difference in the standard deviations a false alarm?

R/checks.R
tjburch: Latest commit covers this

tjburch:
Thanks for handling so many of those @topepo. I think we're up to date but if anything is still lingering, let me know

topepo:
Thanks again!

PR #1008, opened 2025-05-28
Melodie import

topepo:
Brings over new functions and unit tests parallel to what is already in tune


R/loopy.R
```diff
@@ -0,0 +1,286 @@
+loopy <- function(resamples, grid, static) {
```
topepo, line 1: We exported this in melodie, but that was for convenience. Any reason to export it?

DESCRIPTION
```diff
@@ -49,23 +50,22 @@ Suggests:
...
 Remotes:
     tidymodels/hardhat,
     tidymodels/parsnip,
-    tidymodels/recipes,
     tidymodels/rsample,
     tidymodels/tailor,
-    tidymodels/workflows,
-    tidymodels/dials,
-    tidymodels/probably
+    tidymodels/workflows
```
topepo, line 68: I'm loath to do this but it's probably for the best

R/loopy.R
```diff
@@ -0,0 +1,286 @@
...
+            # No postprocessor so just use what we have
+            final_pred <- dplyr::bind_cols(current_pred, current_predict_grid)
+          }
+
+          # --------------------------------------------------------------------
+          # Allocate predictions to an overall object
+
+          pred_iter <- pred_iter + 1
+          # cli::cli_inform("-- -- Allocation {pred_iter} of {nrow(grid)}")
+          # TODO We might not be able to predict ahead of time how many rows should
```
topepo, line 199: It is possible that this part could be the source of the slight slow-down

DESCRIPTION
hfrick: What do you mean here? :) The comment is on removing remotes but that doesn't seem loath-worthy?

R/tune_grid_loop_new.R
```diff
@@ -0,0 +1,206 @@
...
+    # TODO go back and revaluate if the helper functions to schedule certain
+    # stages can be used at this level of the computations (see: Frick(2025))
+
+    # If multiple resamples but preprocessing is cheap (or just a validation set).
+    # Loop over grid rows and splits
+    candidates <- get_row_wise_grid(workflow, grid)
+    # Break all combinations of resamples and candidates into a list of integers
+    # for each combination.
+    inds <- tidyr::crossing(s = seq_along(candidates), b = seq_along(resamples))
+    inds <- vec_list_rowwise(inds)
```
hfrick, line 79: Should this `inds` go into `par_opt`?

R/tune_grid_loop_new.R
```diff
@@ -0,0 +1,206 @@
...
+
+    cl <- loop_call(control, par_opt)
+    res <- rlang::eval_bare(cl)
```
hfrick: Both the if block and the else block make the loop call and then evaluate it. Can we do the modification of `par_opt` in the if statement and move those two lines outside the loop?

R/tune_grid_loop_new.R
```diff
@@ -0,0 +1,206 @@
...
+  # Control execution
+
+  if (control$parallel_over == "resamples") {
+    # The default: Loop over splits, process whole grid
+
+    cl <- loop_call(control, par_opt)
+    res <- rlang::eval_bare(cl)
+  } else {
+    # TODO go back and revaluate if the helper functions to schedule certain
+    # stages can be used at this level of the computations (see: Frick(2025))
```
hfrick: This TODO refers to what is now `get_row_wise_grid()`, let's move it there.

R/loopy.R
hfrick: I currently don't see a reason :) 

Is loopy intented to be the final name? I think it's cute but also a bit generic. What about something along the lines of `loop_over_all_stages()`?

R/loopy.R
```diff
@@ -0,0 +1,286 @@
...
+    dplyr::select(dplyr::starts_with("id"))
+
+  pred_reserve <- NULL
+  pred_iter <- 0
+  notes <- tibble::tibble(
+    location = character(),
+    type = character(),
+    note = character()
+  )
+  # TODO add extras and notes
```
hfrick: Is this front and center in your minds or should this be an issue?

R/tune_grid_loop_new.R
```diff
@@ -0,0 +1,206 @@
...
+      )
+    }
+  }
+  cl
+}
+
+# `new_tune_results()` is exported so let's make a different one
+
+tune_contructor <- function(
+  resamples,
```
hfrick: `new_tune_results()` is not exported but now that we are in tune, we can use it instead of this copy.

DESCRIPTION
topepo: Hmm. That was in err. I thought I had to put a survival dependency in but don't see it now.

R/loopy.R
topepo: Notes are done, extras are not. #1009

R/loopy-helpers.R
```diff
@@ -0,0 +1,524 @@
+# Helpers for loopy()
+#' @import rlang
```
EmilHvitfeldt:
```suggestion
```

R/loopy-helpers.R
```diff
@@ -0,0 +1,524 @@
...
+
+# TODO remove this when merged in to tune
+
+# ------------------------------------------------------------------------------
+
+# Note: in loop(), we add more elements for the outcome name(s(), and the
```
EmilHvitfeldt:
```suggestion
# Note: in loop(), we add more elements for the outcome name(s), and the
```

R/loopy.R
```diff
@@ -0,0 +1,286 @@
...
+    hardhat::extract_spec_parsnip() |>
+    get_submodel_info() |>
+    dplyr::filter(has_submodel) |>
+    purrr::pluck("id")
+
+  const_param <- setdiff(param_tuned, submodel)
+  const_param <- rlang::syms(const_param)
+
+  if (length(submodel) == 0) {
+    inds <- 1:nrow(grid)
```
EmilHvitfeldt:
```suggestion
    inds <- seq_len(nrow(grid))
```

R/tune_grid_loop_new.R
```diff
@@ -0,0 +1,206 @@
...
+  static <- make_static(
+    workflow,
+    param_info = param_info,
+    metrics = metrics,
+    eval_time = eval_time,
+    split_args = split_args,
+    control = control
+  )
+
+  tm_pkgs <- c("rsample", "workflows", "hardhat", "tune", "parsnip", "tailor")
```
EmilHvitfeldt: wouldn't we need yardstick too?

R/tune_grid_loop_new.R
```diff
@@ -0,0 +1,206 @@
...
+    rset_info = rset_info,
+    workflow = workflow
+  )
+}
+
+vec_list_rowwise <- function(x) {
+  vctrs::vec_split(x, by = 1:nrow(x))$val
+}
+
+check_parallel_over <- function(control, resamples) {
```
EmilHvitfeldt: could we change this function name to something like `update_parallel_over()`? as it doesn't really check anything

EmilHvitfeldt (approved):
I agree with Hannah that we should look over all the TODOs and convert to issues if they are still applicable

R/loopy.R
topepo: I will make that change but do it in the next PR after this is merged.

R/tune_grid_loop_new.R
topepo: No. We're making a call that uses `inds` and then evaluating it in the current environment. For example, here's a value of `cl`: 

```r
future.apply::future_lapply(X = inds, FUN = "loopy2", resamples, 
    candidates, static, future.label = "tune-grid-%d", future.stdout = TRUE, 
    future.seed = TRUE, future.packages = load_pkgs)
```

Similarly, for line 58, `resamples` is in `cl`.

R/tune_grid_loop_new.R
topepo: We don't need to alter `par_opt`, but we can consolidate. Changes are in 2111c61670daacd4ef49b67fe3fc09a1c07148a5 along with some comments about the situation (since it's not really obvious to either of us on code examination :smile:).

PR #1027, opened 2025-06-30
updated helpers for parallel processing

topepo:
Some basic helper functions to determine when and how to use parallel processing. 

Docs are updated too. 

The next step is to update [`loop_call()`](https://github.com/tidymodels/tune/blob/3b5fc7c38ea79af76acc64a251ea97839970afa2/R/tune_grid_loop_new.R#L141-L186) with the additional logic. I'll make updates to this branch for that.


R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
+# This are new changes for parallel process via future _or_ mirai. These
+# functions should be able to replace what is contained in `parallel.R`
+
+# ------------------------------------------------------------------------------
+
+has_non_par_pkgs <- function(object, control, verbose = FALSE) {
```
topepo, line 6: I don't think that we'll expose `verbose` to users but we might want it internally.

R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
...
+    pkgs <- required_pkgs(object)
+  }
+  if (!is.null(control)) {
+    pkgs <- c(pkgs, control$pkgs)
+  }
+  pkgs <- unique(pkgs)
+  if (length(pkgs) == 0) {
+    return(FALSE)
+  }
+  naughty_list <- c("keras", "rJava")
```
topepo, line 19: Do we know of any others?

tests/testthat/test-parallel.R
```diff
@@ -0,0 +1,132 @@
...
+  expect_equal(tune:::choose_framework(control = ctrl_java), "sequential")
+  expect_snapshot(tune:::choose_framework(control = ctrl_java, verbose = TRUE))
+
+  daemons(1)
+
+  expect_equal(tune:::choose_framework(), "sequential")
+  expect_snapshot(tune:::choose_framework(verbose = TRUE))
+
+  daemons(0)
+  ## Give daemons a chance to shutdown
```
topepo, line 102: I was seeing issues where some workers were still up and around when resetting `daemons`. I was getting errors `daemons already set for 'default' compute profile`. 

I found this in the future.mirai repo. 

Is there a better way to teardown the workers?

tests/testthat/test-parallel.R
```diff
@@ -0,0 +1,132 @@
...
+
+  # ----------------------------------------------------------------------------
+  # Due to
+
+  expect_equal(tune:::choose_framework(control = ctrl_no), "sequential")
+  expect_equal(tune:::choose_framework(svm_spec), "sequential")
+})
+
+test_that("enable future parallelism", {
+  skip_if_not_installed("future")
```
topepo, line 20: I think we will end up moving future to Suggests so I added this here.

PR #1028, opened 2025-06-30
updated helpers for parallel processing

topepo:
Some basic helper functions to determine when and how to use parallel processing. 

Docs are updated too. 

The next step is to update [`loop_call()`](https://github.com/tidymodels/tune/blob/3b5fc7c38ea79af76acc64a251ea97839970afa2/R/tune_grid_loop_new.R#L141-L186) with the additional logic. I'll make updates to this branch for that.


R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
+# This are new changes for parallel process via future _or_ mirai. These
+# functions should be able to replace what is contained in `parallel.R`
+
+# ------------------------------------------------------------------------------
+
+has_non_par_pkgs <- function(object, control, verbose = FALSE) {
```
topepo, line 6: I don't think that we'll expose `verbose` to users but we might want it internally.

R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
...
+    pkgs <- required_pkgs(object)
+  }
+  if (!is.null(control)) {
+    pkgs <- c(pkgs, control$pkgs)
+  }
+  pkgs <- unique(pkgs)
+  if (length(pkgs) == 0) {
+    return(FALSE)
+  }
+  naughty_list <- c("keras", "rJava")
```
topepo, line 19: Do we know of any others?

tests/testthat/test-parallel.R
```diff
@@ -0,0 +1,132 @@
...
+  expect_equal(tune:::choose_framework(control = ctrl_java), "sequential")
+  expect_snapshot(tune:::choose_framework(control = ctrl_java, verbose = TRUE))
+
+  daemons(1)
+
+  expect_equal(tune:::choose_framework(), "sequential")
+  expect_snapshot(tune:::choose_framework(verbose = TRUE))
+
+  daemons(0)
+  ## Give daemons a chance to shutdown
```
topepo, line 104: I was seeing issues where some workers were still up and around when resetting `daemons`. I was getting errors `daemons already set for 'default' compute profile`. 

I found this in the future.mirai repo. 

Is there a better way to teardown the workers?

tests/testthat/test-parallel.R
```diff
@@ -0,0 +1,132 @@
...
+
+  # ----------------------------------------------------------------------------
+  # Due to
+
+  expect_equal(tune:::choose_framework(control = ctrl_no), "sequential")
+  expect_equal(tune:::choose_framework(svm_spec), "sequential")
+})
+
+test_that("enable future parallelism", {
+  skip_if_not_installed("future")
```
topepo, line 20: I think we will end up moving future to Suggests so I added this here.

topepo:
We are in the middle of major surgery with this repo so I would concentrate on these two files: 

 - [parallel_new.R](https://github.com/tidymodels/tune/pull/1028/files#diff-4a5a81743854e3ab00ba44997f49b521b5975ddf484bcb44b50467c2d34cfa63)
 - [test-parallel.R](https://github.com/tidymodels/tune/pull/1028/files#diff-da826352c208156a64434958b0d3891aaea4edf7d72911ab7f2836931bb4f506)

NEWS.md
```diff
@@ -4,6 +4,8 @@
 
 * Post-processing: new `schedule_grid()` for scheduling a grid including post-processing (#988).
 
+* Updated `?parallelism` to describe changed for future and mirai.
```
simonpcouch:
```suggestion
* Introduced support for parallel processing with mirai in addition to the currently supported framework future. See `?parallelism` to learn more (#1028).
```

R/parallel_new.R
simonpcouch: Seems reasonable to me; this lines up with the [current list](https://github.com/tidymodels/tune/blob/3b5fc7c38ea79af76acc64a251ea97839970afa2/R/parallel.R#L9).

R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
...
+      }
+    }
+  } else {
+    if (verbose) {
+      cli::cli_inform("{.pkg mirai} is not loaded")
+    }
+    mirai_workers <- 0L
+  }
+  mirai_workers
+}
```
simonpcouch: This control flow feels pretty complicated. Maybe:
```suggestion
get_mirai_workers <- function(verbose) {
 if (!mirai_avail()) {
   if (verbose) {
     cli::cli_inform("{.pkg mirai} is not loaded")
   }
   return(0L)
 }
 
 # note connections will be 0 if `!daemons_set()`
 mirai_workers <- mirai::status()$connections
 
 if (verbose) {
   cli::cli_inform(
     "{.pkg mirai} is loaded with {mirai_workers} worker{?s}"
   )
 }
 
 mirai_workers
}
```

simonpcouch: Same early return could simplify the control flow for future

R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
...
+
+  mirai_workers <- get_mirai_workers(verbose)
+  future_workers <- get_future_workers(verbose)
+
+  neither <- future_workers < 2 & mirai_workers < 2
+  both <- future_workers >= 2 & mirai_workers >= 2
+
+  if (neither) {
+    res <- "sequential"
+  } else if (both) {
```
simonpcouch:
```suggestion
  if (neither) {
    return("sequential")
  } 
  if (both) {
```

R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
...
+    if (verbose) {
+      cli::cli_inform(
+        "Multiple workers exist for both {.pkg mirai} and {.pkg future}; 
+        falling back to the default of {.pkg {default}}."
+      )
+    }
+    res <- default
+  } else {
```
simonpcouch:
```suggestion
    return(default)
  }
```

R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
...
+    if (future_workers >= 2) {
+      res <- "future"
+    } else {
+      res <- "mirai"
+    }
+  }
+  res
```
simonpcouch:
```suggestion
  if (future_workers >= 2) {
    return("future")
  }
  "mirai"
```

R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
...
+#' and load the \pkg{future.mirai} package.
+#'
+#' ## Exceptions
+#'
+#' There are a few times when you might specify that you wish to use parallel
+#' processing, but it will revert to sequential execution.
+#' - Many of the control functions (e.g. [control_grid()]) have an argument
+#' called `allow_par`. If this is set to `FALSE`, parallel backends will always
+#' be ignored.
+#' - Some packages, such as \pkg{rJava} and \pkg{keras} are not compatable with
```
simonpcouch:
```suggestion
#' - Some packages, such as \pkg{rJava} and \pkg{keras} are not compatible with
```

R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
...
+#' explict parallelization. If any of these packages are used, sequential
```
simonpcouch:
```suggestion
#' explicit parallelization. If any of these packages are used, sequential
```

R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
...
+  if (length(pkgs) == 0) {
+    return(FALSE)
+  }
+  naughty_list <- c("keras", "rJava")
+  has_pkg <- pkgs %in% naughty_list
+  if (any(has_pkg)) {
+    pkgs <- pkgs[has_pkg]
+    if (verbose) {
+      cli::cli_inform(
+        "These packages cannot be used with explicit parallel processing: {.pkg {pkgs}}."
```
simonpcouch, line 25: I see "explicit" here and one more time in some docs. When I was last hanging out in tune, I don't think I had seen us use this word before; if this term is new as of this PR, I would define it here or remove it

R/parallel_new.R
simonpcouch: A few more attempts to simplify the control flow in this function with early returns; apologies if any are malformed

tests/testthat/test-parallel.R
simonpcouch: 🤷

tests/testthat/test-parallel.R
```diff
@@ -0,0 +1,132 @@
...
+  daemons(1)
+
+  expect_equal(tune:::choose_framework(), "sequential")
+  expect_snapshot(tune:::choose_framework(verbose = TRUE))
+
+  daemons(0)
+  ## Give daemons a chance to shutdown
+  Sys.sleep(10)
+
+  detach("package:mirai", character.only = TRUE)
```
simonpcouch: This would detach the package even if it had been loaded previously. Maybe just namespace the needed functions?

R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
...
+}
+
+# ------------------------------------------------------------------------------
+
+future_avail <- function() {
+  any(search() == "package:future")
+}
+mirai_avail <- function() {
+  any(search() == "package:mirai")
+}
```
simonpcouch, line 39: It's possible to set daemons without loading the mirai package, and same to plan without future loaded. Could the logic here just be based on the information in `future::nbrOfWorkers()` and `mirai::status()`? Checking whether the packages are loaded feels like an unusual mechanism.

tests/testthat/test-parallel.R
shikokuchuo: Ah, you can't just go from `daemons(2)` to `daemons(1)` without resetting first.

The design of mirai is such that once you have daemons set up, you can add / remove daemons at any time.
- Adding (for local daemons) is just a simple call to `launch_local()`
- Removing - there's currently no dedicated function, rather you would set timeouts/task-outs when they're launched - see the arguments to `daemon()`, or you could simply submit `mirai(q())`

tests/testthat/test-parallel.R
```diff
@@ -0,0 +1,132 @@
...
+  # ----------------------------------------------------------------------------
+  # sequential due to restrictions
+
+  expect_equal(tune:::choose_framework(control = ctrl_no), "sequential")
+  expect_snapshot(tune:::choose_framework(control = ctrl_no, verbose = TRUE))
+
+  expect_equal(tune:::choose_framework(control = ctrl_java), "sequential")
+  expect_snapshot(tune:::choose_framework(control = ctrl_java, verbose = TRUE))
+
+  daemons(1)
```
shikokuchuo, line 97:
```suggestion
  daemons(0)
  daemons(1)
```
Need to reset first.

R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
...
+#' @description
+#' There are two frameworks that can be used to parallel process your work: the
+#' [future][future::future] package and the [mirai][mirai:: mirai] package.
+#' Previously, you could use the [foreach][foreach::foreach] package, but this
+#' has been deprecated as of version 1.2.1 of tune.
+#'
+#' By default, no parallelism is used to process models in \pkg{tune}; you have
+#' to opt-in.
+#'
+#' ## Using futures
```
shikokuchuo:
```suggestion
#' ## Using future
```
The package, not the generic term 🙂

R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
...
+#'    library(doBackend)
+#'    registerDoBackend(cores = 4)
+#' ```
+#'
+#' See [future::plan()] for possible options other than `multisession`.
+#'
+#' Note that \pkg{tune} resets the _maximum_ limit of memory of global variables
+#' (e.g., attached packages) to be greater than the default when the package is
+#' loaded. This value can be altered using `options(future.globals.maxSize)`.
+#'
```
shikokuchuo, line 217:
```suggestion
#'
#' If you want \pkg{future} to use \pkg{mirai} parallel workers, you can
#' install and load the \pkg{future.mirai} package.
#' 
```
This is the specific reason you'd want to use future.mirai.

R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
...
+#' ## Using mirai
+#'
+#' To set the specific for parallel processing with \pkg{mirai}, the
+#' [mirai::daemons()] functions. The first argument, `n`, determines the number
+#' of parallel workers. Using `daemons(0)` reverts to sequential processing.
+#'
+#' If you want to use \pkg{future} and \pkg{mirai} together, you can install
+#' and load the \pkg{future.mirai} package.
```
shikokuchuo:
```suggestion
```
Moved to under future.

tests/testthat/test-parallel.R
```diff
@@ -0,0 +1,132 @@
...
+
+  library(future)
+  plan(multisession(workers = 2))
+
+  expect_equal(tune:::choose_framework(), "mirai")
+  expect_snapshot(tune:::choose_framework(verbose = TRUE))
+
+  daemons(0)
+  ## Give daemons a chance to shutdown
+  Sys.sleep(10)
```
shikokuchuo:
```suggestion
  Sys.sleep(1)
```
The 1 second sleep is just safety for CRAN, and not strictly needed here.

tests/testthat/test-parallel.R
shikokuchuo: Either would work for mirai.

R/parallel_new.R
```diff
@@ -0,0 +1,205 @@
...
+#' Note that \pkg{tune} resets the _maximum_ limit of memory of global variables
+#' (e.g., attached packages) to be greater than the default when the package is
+#' loaded. This value can be altered using `options(future.globals.maxSize)`.
+#'
+#' ## Using mirai
+#'
+#' To set the specific for parallel processing with \pkg{mirai}, the
+#' [mirai::daemons()] functions. The first argument, `n`, determines the number
+#' of parallel workers. Using `daemons(0)` reverts to sequential processing.
+#'
```
shikokuchuo, line 226:
```suggestion
#'
#' The arguments `url` and `remote` are used to set up and launch parallel
#' processes over the network for distributed computing. See [mirai::daemons()]
#' documentation for more details.
```

Adds note about distributed computing.

R/parallel_new.R
shikokuchuo: `mirai::status()$connections` is the most general way of getting what you want here, but it does query the status, which is probably not necessary. As I understand it though, this is to choose the framework and isn't called very frequently so we may as well go with what's convenient.

tests/testthat/test-parallel.R
```diff
@@ -0,0 +1,132 @@
...
+  expect_snapshot(tune:::choose_framework(control = ctrl_java, verbose = TRUE))
+
+  daemons(1)
+
+  expect_equal(tune:::choose_framework(), "sequential")
+  expect_snapshot(tune:::choose_framework(verbose = TRUE))
+
+  daemons(0)
+  ## Give daemons a chance to shutdown
+  Sys.sleep(10)
```
shikokuchuo:
```suggestion
  Sys.sleep(1)
```
Just safety for CRAN 1s not 10s!

R/parallel_new.R
topepo: The plan is for both mirai and future to be in suggests so someone will have to install both even if they don't want to run in parallel.

R/parallel_new.R
topepo: I'll write more about it in this doc, but it's a long-standing classification of different parallel processing strategies. The reference that I usually cite is

> Tierney, Luke. "Implicit and explicit parallel computing in R." COMPSTAT 2008: Proceedings in Computational Statistics. Physica-Verlag HD, 2008.

Issue #1038, opened 2025-07-08
better use of random numbers for one-shot work and calibration


topepo:
With the new random number generation scheme we are resetting the RNG stream. That's very helpful with multiple tasks, but for a single task (e.g. `last_fit()`), we lose reproducibility. 

We should be able to detect when we are in this scenario and leave the RNGs alone throughout the run. 

Also, `inner_split()` uses random numbers sometimes; we should also have a way of controlling those (so that we can compare to usage outside of tune). This happens in the middle of the looping structure, so it is more difficult to reproduce outside of tune (but I have some ideas).

PR #1043, opened 2025-07-14
No extra rng seed alterations for last_fit()

topepo:
As mentioned in #1038, we want to ensure that users can reproduce the results of `last_fit()` exactly. 

This PR does not generate (or set) the rng seeds when `last_fit()` is the parent function. Consequently, the random numbers used manually should be the same as those used by last fit **unless there is a stochastic inner split**. 

Additional tests will be in extratests since we don't suggest any modeling packages in tune that we could test with.

### DIFF
--- a/R/parallel_new.R
+++ b/R/parallel_new.R
@@ -268,7 +268,7 @@ get_parallel_seeds <- function(workers) {
 #'
 #' Also, we want the results for [last_fit()] to be the same as what one would
 #' get by executing those steps manually (i.e., training the workflow,
-#' predicting the test set, etc.). If [last_fit()] is used, we don’t affect the
+#' predicting the test set, etc.). If [last_fit()] is used, we don’t alter the
 #' random number stream.
 #'
 #' @references
--- a/R/parallel_new.R
+++ b/R/parallel_new.R
@@ -183,7 +183,7 @@ get_parallel_seeds <- function(workers) {
 #'
 #' @description
 #'
-#' \pkg{tune} can enable the simultaneous parallel computations. Tierney (2008)
+#' \pkg{tune} can enable simultaneous parallel computations. Tierney (2008)
 #'  defined different classes of parallel processing techniques:
 #'
 #'  - _Implicit_ is when a function uses low-level tools to perform a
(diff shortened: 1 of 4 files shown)


R/parallel_new.R
```diff
@@ -266,6 +266,11 @@ get_parallel_seeds <- function(workers) {
 #' - If you specify fewer than two workers, or if there is only a single task,
 #'  the computations will occur sequentially.
 #'
+#' Also, we want the results for [last_fit()] to be the same as what one would
+#' get by executing those steps manually (i.e., training the workflow,
+#' predicting the test set, etc.). If [last_fit()] is used, we don’t affect the
```
simonpcouch:
```suggestion
#' predicting the test set, etc.). If [last_fit()] is used, we don’t alter the
```

nit

simonpcouch (commented):
My only concern had been whether "leftover" workers from other work done in parallel might trigger an unintended condition with RNG, but all good:

``` r
# setup
library(tidymodels)

set.seed(1)
dat <- sim_regression(200)
split <- initial_split(dat)
tr_dat <- training(split)
te_dat <- testing(split)
rs <- vfold_cv(tr_dat)

rf_spec <- rand_forest(mode = "regression", trees = 20)
rf_wflow <- workflow(outcome ~ ., rf_spec)

# manual:
set.seed(2)
manual_fit <- fit(rf_wflow, tr_dat)
manual_pred <- 
  augment(manual_fit, new_data = te_dat) |> 
  select(outcome, .pred)

# auto, sequential:
set.seed(2)
auto_res <- last_fit(rf_wflow, split)
auto_pred <- 
  auto_res |> 
  collect_predictions() |> 
  select(outcome, .pred)

# auto, "parallel"
library(mirai)
daemons(5)
#> [1] 5

set.seed(2)
auto_res_par <- last_fit(rf_wflow, split)
auto_pred_par <- 
  auto_res |> 
  collect_predictions() |> 
  select(outcome, .pred)

# compare
waldo::compare(manual_pred, auto_pred)
#> ✔ No differences
waldo::compare(auto_pred, auto_pred_par)
#> ✔ No differences
```

<sup>Created on 2025-07-15 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

LGTM, but leaving reviewing as Comment since I'm not very familiar with the codebase at this point.

R/parallel_new.R
```diff
@@ -266,6 +266,11 @@ get_parallel_seeds <- function(workers) {
 #' - If you specify fewer than two workers, or if there is only a single task,
 #'  the computations will occur sequentially.
 #'
+#' Also, we want the results for [last_fit()] to be the same as what one would
+#' get by executing those steps manually (i.e., training the workflow,
+#' predicting the test set, etc.). If [last_fit()] is used, we don’t alter the
+#' random number stream.
```
hfrick: The reference to the random number stream is a lot more technical than the rest of the info here and feels a little out of place to me. Given that  `last_fit()` essentially has only one task (one workflow fitted to one training set), it follows from the 3rd exception above that it's carried out sequentially, so things should be reproducible. 

Talking about what is reproducible and when, is a good thing though! I would just make this its own section, and add a note about running sequentially vs in parallel (and that it doesn't matter if future or mirai), as well as tune 1.3.0 vs new tune.

hfrick (commented):
LGTM! The new text on parallelism reads nicely. The first sentence on mirai is missing something, though. I just can't comment on it directly because Github things it's unrelated to the PR.

R/parallel_new.R
topepo: Emphasis added by me: 

> it follows from the 3rd exception above that it's carried out sequentially, so **things should be reproducible**.

The problem was that `last_fit()` wasn't reproducible in a different way. We're fixing that here. 

I've reworded some things, removed references to the random number stream, and added a few more specific exceptions that I could think of.

hfrick: The new section is really clear! 😍

PR #1058, opened 2025-07-26
move over melodie logging to tune logging

EmilHvitfeldt:


NAMESPACE
```diff
@@ -135,7 +135,6 @@ S3method(vec_restore,iteration_results)
 S3method(vec_restore,resample_results)
 S3method(vec_restore,tune_results)
 export(.catch_and_log)
-export(.catch_and_log_fit)
```
EmilHvitfeldt, line 138: This new way doesn't appear to need this.

It is only used in tune
https://github.com/search?q=org%3Acran%20tune%20.catch_and_log_fit&type=code

and we document it as internal so i feel this is safe

tests/testthat/test-logging.R
```diff
@@ -25,109 +25,6 @@ test_that("tune_log", {
   expect_snapshot(tune_log(ctrl_t, rs, task = "cube", type = "success"))
 })
 
-test_that("log issues", {
```
EmilHvitfeldt, line 28: There were a few differences in how we did things, which made these tests unnecessary or needlessly complicated to replicate. I'm deleting since these functions are tested by proxy later

topepo (approved):
Looks great. Thanks!

PR #1059, opened 2025-07-27
Merge "melodie" changes into main

topepo:
These changes are a systematic rewrite of the underlying tools that power `tune_grid()`. They were updated so that we can include postprocessing and also our final farewell to the foreach package (in favor of future and/or mirai). 

One important note is that the pattern used for the `.config` column in the results will have a different pattern.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -31,7 +31,7 @@ Imports:
     purrr (>= 1.0.0),
     recipes (>= 1.1.0.9001),
     rlang (>= 1.1.4),
-    rsample (>= 1.3.0.9002),
+    rsample (>= 1.3.0.9003),
     tailor,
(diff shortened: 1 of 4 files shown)


topepo:
@EmilHvitfeldt I am seeing this warning: 

```
N  checking R code for possible problems (8.1s)
   log_catalog: no visible global function definition for ‘tune_catalog’
   Undefined global functions or variables:
     tune_catalog
```

We use (but don't define) `tune_catalog()` on logging.R:184. That is called inside `log_catalog()`, which is called from `tune_log()`.

EmilHvitfeldt:
That appears to be some leftover code i didn't catch as i was only doing `devtools::test()`, should be fixed now

PR #1060, opened 2025-07-28
use new accessor function `calibration()`

hfrick:
This PR keeps `melodie-switch` in sync with rsample https://github.com/tidymodels/rsample/pull/586


topepo (commented):
Perfecto

PR #1065, opened 2025-07-29
Load required packages in workers

topepo:
When adding Mirai, we left out the bits where we attach the required packages. This PR does that in the loops and leaves some infrastructure for cases where a package will only function well when fully loaded (maybe earth is an example).


topepo:
🙄 Need to redo since the merge to `main`

PR #1066, opened 2025-07-29
Load required packages in workers

topepo:
When adding Mirai, we left out the bits where we attach the required packages. This PR does that in the loops and leaves some infrastructure for cases where a package will only function well when fully loaded (maybe earth is an example).

### DIFF
--- a/R/loop_over_all_stages-helpers.R
+++ b/R/loop_over_all_stages-helpers.R
@@ -11,6 +11,7 @@ make_static <- function(
   eval_time,
   split_args,
   control,
+  pkgs = character(0),
   data = list(fit = NULL, pred = NULL, cal = NULL)
 ) {
   # check inputs
@@ -40,6 +41,7 @@ make_static <- function(
     eval_time = eval_time,
     split_args = split_args,
     control = control,
+    pkgs = pkgs,
     data = data
   )
 }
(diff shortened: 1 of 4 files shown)


R/loop_over_all_stages.R
```diff
@@ -4,8 +4,11 @@
 #    `allow_par = FALSE`; that will use `lapply()` so that you can see output.
 
 loop_over_all_stages <- function(resamples, grid, static) {
-  # Initialize some objects
+  # Some packages may use random numbers so attach them prior to initializing
+  # the RNG seed
+  attach_pkgs(static$pkgs)
```
EmilHvitfeldt, line 9: `attach_pkgs()` is defined to have 2 arguments `pkgs` and `load`. Only `pkgs` is used, are you planning to use `attach_pkgs()` other places?

R/loop_over_all_stages-helpers.R
```diff
@@ -543,3 +545,22 @@ parsnip_to_engine <- function(wflow, grid) {
...
+
+# ------------------------------------------------------------------------------
+
+attach_pkgs <- function(pkgs, load = character(0)) {
+  # Testing to see if they are installed triggers package attachment.
+  is_inst <- purrr::map_lgl(pkgs, rlang::is_installed)
+  if (any(!is_inst)) {
+    nms <- pkgs[!is_inst]
+    cli::cli_abort("Some package installs are needed: {.pkg {nms}}", call = NULL)
+  }
```
EmilHvitfeldt, line 557: Is there a reason why you aren't using `check_installed()`?

```r
rlang::check_installed(pkgs, call = NULL)
```

R/loop_over_all_stages.R
topepo: No but I'm unsure what to add to the `load` argument. It will require some testing. IIRC earth needs to be fully loaded. It's a placeholder for now.

R/loop_over_all_stages-helpers.R
topepo: I was going to use that, but I thought we should make sure that they have the right installs and still have the effect of attaching them.

PR #1072, opened 2025-08-03
load some packages that users interact with directly

topepo:
Since we are attaching packages in the workers, some functions are not available since the packages are not fully loaded: 

``` r
library(tidymodels)
library(mirai)
daemons(10)
#> [1] 10

data(ames)

ames <- mutate(ames, Sale_Price = log10(Sale_Price))

set.seed(502)
ames_split <- initial_split(ames, prop = 0.80, strata = Sale_Price)
ames_train <- training(ames_split)
ames_test  <-  testing(ames_split)

set.seed(1001)
ames_folds <- vfold_cv(ames_train, v = 10)

reg_mtr <- metric_set(rmse, ccc)

ames_rec <-
  recipe(
    Sale_Price ~ Neighborhood + Gr_Liv_Area + Year_Built + Bldg_Type +
      Latitude + Longitude,
    data = ames_train
  ) %>%
  step_other(Neighborhood, threshold = 0.01) %>%
  step_dummy(all_nominal_predictors()) %>%
  step_interact(~ Gr_Liv_Area:starts_with("Bldg_Type_")) %>%
  step_ns(Latitude, Longitude, deg_free = 20)

lm_wflow <-
  workflow() %>%
  add_recipe(ames_rec) %>%
  add_model(linear_reg() %>% set_engine("lm"))

lm_fit <- lm_wflow %>% fit(data = ames_train)

get_model <- function(x) {
  library(broom)
  extract_fit_parsnip(x) %>% tidy()
}

ctrl <- control_resamples(extract = get_model)

lm_res <-
  lm_wflow %>%
  fit_resamples(resamples = ames_folds, control = ctrl, metrics = reg_mtr)
#> Warning: All models failed. Run `show_notes(.Last.tune.result)` for more
#> information.
show_notes(.Last.tune.result)
#> unique notes:
#> ──────────────────────────────────────────────────
#> Error in `step_dummy()`:
#> Caused by error in `prep()`:
#> Caused by error in `all_nominal_predictors()`:
#> ! could not find function "all_nominal_predictors"
```

<sup>Created on 2025-08-02 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

This PR fully loads the packages that people are most likely to specify.

### DIFF
--- a/R/loop_over_all_stages-helpers.R
+++ b/R/loop_over_all_stages-helpers.R
@@ -4,15 +4,16 @@
 # Note: in loop(), we add more elements for the outcome name(s), and the
 # data partitions
 make_static <- function(
-  workflow,
-  param_info,
-  grid,
-  metrics,
-  eval_time,
-  split_args,
-  control,
-  pkgs = character(0),
-  data = list(fit = NULL, pred = NULL, cal = NULL)
+    workflow,
+    param_info,
+    grid,
+    metrics,
+    eval_time,
+    split_args,
+    control,
+    pkgs = character(0),
+    strategy = character(0),
+    data = list(fit = NULL, pred = NULL, cal = NULL)
 ) {
   # check inputs
   if (!inherits(workflow, "workflow")) {
@@ -42,6 +43,7 @@ make_static <- function(
     split_args = split_args,
     control = control,
     pkgs = pkgs,
+    strategy = strategy,
     data = data
   )
 }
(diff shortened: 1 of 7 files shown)


R/loop_over_all_stages-helpers.R
```diff
@@ -561,6 +561,13 @@ attach_pkgs <- function(pkgs, load = character(0)) {
 
   # There may be some packages that need to be fully loaded to work
   # appropriately.
+
+  # These are packages that users might directly interact with so we should
+  # fully load them
+  load_user_pkgs <- c("parsnip", "recipes", "workflows", "tailor", "tune")
```
topepo: I don't know if this is enough or if we should just fully load everything (which I think is what happened previously)

topepo:
@hfrick I could use a re-review on this

R/parallel.R
```diff
@@ -295,6 +295,17 @@ get_parallel_seeds <- function(workers) {
 #'   platforms due to the packages' use of different numerical tolerance
 #'   constants across operating systems.
 #'
+#' ## Handling package dependencies
+#'
+#' \pkg{tune} knows what packages are required to fit a workflow object.
+#'
+#' When computations are run sequentially, an initial check is made to see if
+#' they are installed. This triggers the packages to be attached (but not fully
+#' loaded).
```
hfrick: From https://r-pkgs.org/dependencies-mindset-background.html#sec-dependencies-attach-vs-load

> It’s common to hear something like “we use [library(somepackage)](https://rdrr.io/r/base/library.html) to load somepackage”. But technically [library()](https://rdrr.io/r/base/library.html) attaches a package to the search path. This casual abuse of terminology is often harmless and can even be beneficial in some settings. But sometimes it’s important to be precise and pedantic and this is one of those times.

I think our situation here is also one where we'd want to be precise and pedantic. 

In the above reference is also 
> You can’t attach a package without first loading it

so I'd change the wording above about attached but not fully loaded.

R/loop_over_all_stages-helpers.R
```diff
@@ -548,21 +550,19 @@ parsnip_to_engine <- function(wflow, grid) {
...
-attach_pkgs <- function(pkgs, load = character(0)) {
-  # Testing to see if they are installed triggers package attachment.
-  is_inst <- purrr::map_lgl(pkgs, rlang::is_installed)
-  if (any(!is_inst)) {
-    nms <- pkgs[!is_inst]
-    cli::cli_abort(
-      "Some package installs are needed: {.pkg {nms}}",
-      call = NULL
-    )
+attach_pkgs <- function(pkgs, strategy = "sequential", load = character(0)) {
```
hfrick: What is the intended difference between `pkgs` and `load`? 

Both sets of packages get attached (and thus also loaded) via `library()` in this function, so we can potentially simplify this function

R/loop_over_all_stages.R
```diff
@@ -6,7 +6,7 @@
 loop_over_all_stages <- function(resamples, grid, static) {
   # Some packages may use random numbers so attach them prior to initializing
   # the RNG seed
-  attach_pkgs(static$pkgs)
+  attach_pkgs(static$pkgs, strategy = static$strategy)
```
hfrick, line 9: We don't make use of the `load` argument to `attach_pkgs()` so another indicator that we might be able to get rid of it

hfrick (commented):
I think this still needs a small tweak and then will fall into place nicely

Issue #1074, opened 2025-08-05
`control$allow_par` and `allow_parallelism()`


EmilHvitfeldt:
I think we need to handle the following assumption in `allow_parallelism()`

https://github.com/tidymodels/tune/blob/03d7cb6baf3cd56303afc8d8601209c3d7d2dba3/R/parallel.R#L453

The following code on CRAN returns `FALSE` for both of these, as it used `foreach::getDoParWorkers() > 1 || future::nbrOfWorkers() > 1` to help determine whether we can run things in parallel.

```r
# Input is control$allow_par
tune:::allow_parallelism(TRUE)
tune:::allow_parallelism(FALSE)
```

On dev right now we don't do that it returns `TRUE` and `FALSE` respectively. The only difference right now is that the logging seen below only shows up when `allow_par = FALSE` which is not the default

<img width="684" height="138" alt="Image" src="https://github.com/user-attachments/assets/7c8e18a6-b51b-4d3c-8920-f8fd4a6ece53" />

topepo:
Related to #1057

We should avoid `allow_parallel()` and use the results of `choose_framework()` instead, with `framework != "sequential"` indicating that we will run in parallel. 

This means that `initialize_catalog()` will need to take the workflow as an argument (along with the existing control argument) to pass along to `choose_framework()`.

PR #1079, opened 2025-08-05
Use choose framework in logging

EmilHvitfeldt:
to close https://github.com/tidymodels/tune/issues/1074

### DIFF
--- a/R/logging.R
+++ b/R/logging.R
@@ -142,7 +142,11 @@ catalog_is_active <- function() {
 #
 #' @rdname tune-internal-functions
 #' @export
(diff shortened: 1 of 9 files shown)


R/logging.R
```diff
@@ -154,9 +158,13 @@ initialize_catalog <- function(control, env = rlang::caller_env()) {
       nrow = 0
     )
 
+  if (is.null(workflow)) {
+    workflow <- workflow()
+  }
```
EmilHvitfeldt, line 163: This is purely added so finetune doesn't break, in the case where you have an old version of finetune and a new version of tune

topepo (approved):
Very nice!

PR #1081, opened 2025-08-11
Smooth out fitting the post stage

hfrick:
Draft stage: need to finalize and merge https://github.com/tidymodels/workflows/pull/298 first (and change the remote from that PR to workflows `main` branch)

tune fits the workflow in stages, so that we can, e.g., do expensive preprocessing only as often as necessary.  For the `pre` and `model` stages, we have `finalize_fit_pre()` and `finalize_fit_model()`, which both extract the relevant component, finalize it, write it back into the workflow, and then use the corresponding `.fit_*()` function from workflows to fit the stage (within the workflow). 

This PR applies the same pattern to the `post` stage, making use of `workflows::.fit_post()`. This means that, for calibration, we do not have to generate predictions on the calibration set manually, `.fit_post()` is taking care of that.

As part of this, this PR also fixes where the unfitted and the fitted versions of the tailor object are stored within the workflow: the unfitted version in the `action`, the fitted version in the `fit` slot.

We could also use `predict.workflow()`. It generates the predictions from the (primary) model and then adjusts them based on the (fitted) tailor. Since we already generated the model predictions earlier in the loop, we extract the fitted tailor and adjust those instead of re-generating them as part of `predict.workflow()`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -39,7 +39,7 @@ Imports:
     tidyselect (>= 1.1.2),
     vctrs (>= 0.6.1),
     withr,
-    workflows (>= 1.2.0.9000),
+    workflows (>= 1.2.0.9001),
     yardstick (>= 1.3.0)
 Suggests: 
     C50,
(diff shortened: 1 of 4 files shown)


DESCRIPTION
```diff
@@ -64,7 +64,7 @@ Suggests:
 Remotes:
     tidymodels/rsample,
     tidymodels/tailor,
-    tidymodels/workflows
+    tidymodels/workflows#298
```
EmilHvitfeldt: putting as a reminder that we need to remove this before merging

PR #1091, opened 2025-08-27
Remove remotes

hfrick:
tailor and workflows are on CRAN


EmilHvitfeldt (approved):
Could you also set min required version of tailor? thank you!

PR #1104, opened 2025-10-17
Switch from GPfit to GauPro for Gaussian Processes

topepo:
The GPfit maintainer contacted me a few months ago about the potential for that package to fall off CRAN. This change helps from a sustainability point of view, but also adds some new features. 

   - Fit diagnostics are computed and reported. If the fit quality is poor, an "uncertainty sample" that is furthest away from the existing data is used as the new candidate. 
   - The GP no longer uses binary indicators for qualitative predictors. Instead, a "categorical kernel" is used for those parameter columns. Fewer starting values are required with this change. 
   - For numeric predictors, the Matern 3/2 kernel is always used. 

I've tested this extensively on the analyses in [aml4td](https://aml4td.org/chapters/iterative-search.html#fig-nnet-sg-iterations) and [TMwR](https://www.tmwr.org/iterative-search#fig:progress-plot) and get very similar results.

### DIFF
--- a/R/0_imports.R
+++ b/R/0_imports.R
@@ -125,7 +125,7 @@ utils::globalVariables(
     "predict_stage",
     "se",
     "num",
-    "user"
+    "user",
     ".resample_weight",
     "effective_n"
   )


inst/test_objects.R
```diff
@@ -1,4 +1,17 @@
-library(tidymodels)
+library(broom)
```
topepo, line 1: Note: we were getting warnings with the old tests that: 

> namespace 'workflowsets' is not available and has been replaced

_Not_ loading tidymodels solves this and it comes up in the tests for #1007 1007

R/checks.R
```diff
@@ -244,10 +244,7 @@ check_bayes_initial_size <- function(num_param, num_grid, race = FALSE) {
     c(
       `!` = "{msg}",
       `*` = cli::pluralize(
-        "There are {cli::qty(diff)}{?as many/more} tuning parameters
-          {cli::qty(diff)}{?as/than} there are initial points.
```
EmilHvitfeldt, line 268: Why are these lines being deleted?

if that is intentional then we also don't need `cli::pluralize()`

R/tune_bayes.R
```diff
@@ -807,6 +728,7 @@ initial_info <- function(stats, metrics, maximize, eval_time) {
     dplyr::filter(.metric == metrics) |>
     dplyr::filter(!is.na(mean))
 
+  # TODO a lot of slice_min/slice_max can be used now
```
EmilHvitfeldt, line 731: Should add as an issue if you are not planning on dealing with it in this PR.

R/gp_helpers.R
```diff
@@ -0,0 +1,347 @@
...
+    msg <- cli::format_inline(
+      "GP has a coverage rate < {round(gp_threshold * 100, 1)}% and is unreliable."
+    )
+    message_wrap(
+      msg,
+      prefix = cli::symbol$checkbox_circle_on,
+      color_text = get_tune_colors()$message$danger
+    )
+  }
+  list(use = !loo_bad & !model_fail & loo_rsq > gp_threshold, rsq = loo_rsq)
```
EmilHvitfeldt:
```suggestion
  list(use = !loo_bad && !model_fail && loo_rsq > gp_threshold, rsq = loo_rsq)
```

R/gp_helpers.R
```diff
@@ -0,0 +1,347 @@
...
+    )
+    message_wrap(
+      msg,
+      prefix = "!",
+      color_text = get_tune_colors()$message$warning
+    )
+  }
+
+  withr::with_seed(
+    114,
```
EmilHvitfeldt: why are we hardcoding this specific seed?

R/gp_helpers.R
```diff
@@ -0,0 +1,347 @@
...
+  dplyr::bind_cols(candidates, gp_pred)
+}
+
+
+pick_candidate <- function(results, info, control) {
+  bad_gp <- all(is.na(results$.mean))
+  if (!bad_gp & info$uncertainty < control$uncertain) {
+    results <- results |>
+      dplyr::arrange(dplyr::desc(objective)) |>
+      dplyr::slice(1)
```
EmilHvitfeldt:
```suggestion
      dplyr::slice_max(objective, n = 1, with_ties = FALSE)
```

R/gp_helpers.R
```diff
@@ -0,0 +1,347 @@
...
+    if (control$verbose_iter) {
+      message_wrap(
+        "Uncertainty sample",
+        prefix = cli::symbol$info,
+        color_text = get_tune_colors()$message$info
+      )
+    }
+    results <- results |>
+      dplyr::arrange(dplyr::desc(.sd)) |>
+      dplyr::slice(1:floor(.1 * nrow(results))) |>
```
EmilHvitfeldt:
```suggestion
      dplyr::slice(seq_len(floor(.1 * nrow(results)))) |>
```

R/gp_helpers.R
```diff
@@ -0,0 +1,347 @@
...
+    message_wrap(
+      "Generating a candidate as far away from existing points as possible.",
+      prefix = cli::symbol$info,
+      color_text = get_tune_colors()$message$info
+    )
+
+    return(candidates)
+  } else {
+    message_wrap(
+      paste("Generating", nrow(candidates), "candidates"),
```
EmilHvitfeldt:
```suggestion
      paste("Generating", nrow(candidates), "candidates."),
```

R/checks.R
topepo: We have this warning now. which I find more informative. I don't think that we need both. 

```r
    msg <- cli::format_inline(
      "The Gaussian process model is being fit using {num_pred} feature{?s} but
			only has {num_cand} data point{?s} to do so. This may cause errors or a
			poor model fit."
    )
```

R/gp_helpers.R
topepo: That is a leftover from development. I've taken it out.

PR #1105, opened 2025-10-17
Workflow size warning threshold

### LINKED ISSUE #914: Add a `check_workflow_size` argument to the control functions
[This check](https://github.com/tidymodels/tune/blob/35464b959ebb4b82a15f694d2e6c9fb252489416/R/tune_grid.R#L415) can be a little annoying and we have an arbitrary cutoff. This would make an argument for the cutoff.

topepo:
Closes #914 

Loooong awaited

### DIFF
--- a/R/control.R
+++ b/R/control.R
@@ -203,8 +203,8 @@ print.control_last_fit <- function(x, ...) {
 #'   backend. Defaults to `NULL` for default backend options.
 #' @param allow_par A logical to allow parallel processing (if a parallel
 #'   backend is registered).
-#' @param workflow_size A non-negative number that is used as a threshold for a
-#' warning regarding the size of the workflow. Only used when
+#' @param workflow_size A non-negative number (in MB) that is used as a
+#' threshold for a warning regarding the size of the workflow. Only used when
 #' `save_workflow = TRUE`.
 #'
 #' @inheritSection collect_predictions Hyperparameters and extracted objects
--- a/R/tune_grid.R
+++ b/R/tune_grid.R
@@ -436,16 +436,19 @@ pull_rset_attributes <- function(x) {
 
 set_workflow <- function(workflow, control) {
   if (control$save_workflow) {
-    if (!is.null(workflow$pre$actions$recipe)) {
-      w_size <- utils::object.size(workflow$pre$actions$recipe)
-      # make 5MB cutoff
+    if (!is.null(workflow)) {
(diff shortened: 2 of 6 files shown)


R/control.R
```diff
@@ -200,6 +203,9 @@ print.control_last_fit <- function(x, ...) {
 #'   backend. Defaults to `NULL` for default backend options.
 #' @param allow_par A logical to allow parallel processing (if a parallel
 #'   backend is registered).
+#' @param workflow_size A non-negative number that is used as a threshold for a
+#' warning regarding the size of the workflow. Only used when
+#' `save_workflow = TRUE`.
```
EmilHvitfeldt: it should list that the number is read as MB and that the default value is 100MB

R/tune_grid.R
```diff
@@ -439,7 +439,7 @@ set_workflow <- function(workflow, control) {
     if (!is.null(workflow$pre$actions$recipe)) {
       w_size <- utils::object.size(workflow$pre$actions$recipe)
       # make 5MB cutoff
```
EmilHvitfeldt: should be updated or removed. properly removed

R/control.R
EmilHvitfeldt: also, i'm noticing that the documentation and news is referring to workflow size, however we are only measuring recipes size.

I feel like we should do

```r
utils::object.size(workflow)
```

instead of

```r
utils::object.size(workflow$pre$actions$recipe)
```

tests/testthat/_snaps/control.md
```diff
@@ -0,0 +1,11 @@
+# workflow size warning
+
+    Code
+      set.seed(1)
+      warns <- fit_resamples(lm_wflow, resamples = vfold_cv(MTCARS), control = control_resamples(
+        save_workflow = TRUE, workflow_size = 2))
+    Message
+      i The workflow being saved contains a recipe, which is 2.7 Mb in i memory. If
```
EmilHvitfeldt: should the message include the threshold and where it is set?

EmilHvitfeldt (approved):
This seems like a nice improvement!

what is the rational from changing the default from 5 to 100?

topepo:
> what is the rational from changing the default from 5 to 100?

There really is no right answer. 

5MB is objectively too low. 100MB is subjectively just right (based on my prior). 

Recommendations are welcome

PR #1108, opened 2025-10-26
Add print methods for control objects with formatted output using cli

### LINKED ISSUE #1051: Better print method for control objects
Very low priority.

The print methods doesn't really say much, making them not that useful. I was thinking we could have them print the settings that are different than defaults. Or just all settings would be fine too.

And they should be switched to {cli}

``` r
library(tune)

control_grid()
#> grid/resamples control object

control_grid(verbose = TRUE)
#> grid/resamples control object
```

<sup>Created on 2025-07-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

tjburch:
Closes #1051 by adding a cli-based printout of control objects. Went pretty simple, just a cli bulleted list of all parameters. Doing all is up for debate - I figure it's a bit nicer in case users don't know defaults off the cuff, but can see the argument for only non-default values, the prints do get a little verbose.

```
r$> control_grid()
Grid/resamples control object
  verbose: FALSE
  allow_par: TRUE
  extract: NULL
  save_pred: FALSE
  pkgs: NULL
  save_workflow: FALSE
  event_level: 'first'
  parallel_over: NULL
  backend_options: NULL
  workflow_size: 100
```

```
r$> control_bayes(
      verbose_iter = TRUE,
      no_improve = 10,
      uncertain = 5
    )
Bayes control object
  verbose: FALSE
  verbose_iter: TRUE
  allow_par: TRUE
  no_improve: 10
  uncertain: 5
  seed: 9642
  extract: NULL
  save_pred: FALSE
  time_limit: NA
  pkgs: NULL
  save_workflow: FALSE
  save_gp_scoring: FALSE
  event_level: 'first'
  parallel_over: NULL
  backend_options: NULL
  workflow_size: 100
```

```
$> control_last_fit(
      verbose = TRUE,
      event_level = "second",
      allow_par = TRUE
    )
Last fit control object
  verbose: TRUE
  allow_par: TRUE
  extract: <function>
  save_pred: TRUE
  pkgs: NULL
  save_workflow: FALSE
  event_level: 'second'
  parallel_over: NULL
  backend_options: NULL
  workflow_size: 100
```


tjburch:
Thanks - implemented. The <function> print is still kind of ugly, but the rest now use the cli classes better.

tests/testthat/test-checks.R
```diff
@@ -395,6 +395,39 @@ test_that("Bayes control objects", {
   expect_s3_class(control_bayes(), "control_bayes")
 })
 
+test_that("control object print methods", {
```
topepo: I think that this should live in `test-control.R`

R/control.R
```diff
@@ -72,9 +72,43 @@ control_grid <- function(
...
+
+  # Build formatted lines for each field
+  for (field in fields) {
+    value <- x[[field]]
+
+    if (is.function(value)) {
+      cli::cli_bullets(c(" " = "{.arg {field}}: <function>"))
+    } else if (inherits(value, "tune_backend_options")) {
+      cli::cli_bullets(c(" " = "{.arg {field}}: <backend_options>"))
+    } else {
```
topepo, line 103: Can we add a branch for NULL values before the last that is: 

```r
 else if (is.null(value)) {
      cli::cli_bullets(c(" " = "{.arg {field}}: NULL"))
    }
```

We can then get the output without the empty spaces: 

```
Grid/resamples control object
  `verbose`: FALSE
  `allow_par`: TRUE
  `extract`: NULL
  `save_pred`: FALSE
  `pkgs`: NULL
  `save_workflow`: FALSE
  `event_level`: "first"
  `parallel_over`: NULL
  `backend_options`: NULL
  `workflow_size`: 100
```

topepo:
I thought that we could do something fancy with `formals()` but this works nicely.

R/control.R
tjburch: Done in latest commit

tjburch:
Cool, just implemented NULL handling in the latest commit and put tests in the right place. Let me know if there's anything else.

Ignore the claude reference above - got some free credits dumped into my account that I figured I'd see what it could do and didn't realize it would go off and put its changes on git.

PR #1133, opened 2026-02-04
make work with ordered_prob_metric

EmilHvitfeldt:
These changes makes it so {tune} works with the new ordered prob metrics (`ranked_prob_score()`) from {yardstick}.

As far as I can tell we don't need any version requirements. We are not using any new functions from the new version of {yardstick} so I think we are good.


topepo (commented):
11/10 no notes

PR #1141, opened 2026-03-05
tidyclust loopy integration

EmilHvitfeldt:
This PR is the first push to have tidyclust leverage the new tuning infrastructure. 

## Note on Parameter Registries

- **parsnip models**: Use `parsnip::get_from_env()` which queries parsnip's internal environment
- **tidyclust models**: Use modelenv for registration, but parsnip doesn't see it
- **Solution**: `get_param_info()` handles NULL gracefully, defaulting `has_submodel = FALSE`

alternative solution is to import modelenv and have it swap between them depending on the model type. Also worth noting that tidyclust doesn't have any models that use the submodel trick.

This means tidyclust models won't benefit from submodel optimization (multipredict), but this is a minor performance consideration.


R/loop_over_all_stages.R
```diff
@@ -28,8 +28,11 @@
   split <- resamples$splits[[1]]
   split_labs <- resamples |>
     dplyr::select(dplyr::starts_with("id"))
+  # Determine if we're using prediction-based or model-based metrics
+  is_model_metrics <- identical(static$metric_type, "model")
 
   pred_reserve <- NULL
+  metrics_reserve <- NULL
```
EmilHvitfeldt, line 35: We need to allocate it here since clustering means the relevant code don't run

R/loop_over_all_stages-helpers.R
```diff
@@ -50,6 +69,18 @@ make_static <- function(
   )
 }
 
+is_valid_metric_set <- function(x) {
+  inherits(x, "metric_set") || inherits(x, "cluster_metric_set")
```
hfrick: Would it make sense for `"cluster_metric_set"` to be a subclass of `"metric_set"`?

R/finalize.R
```diff
@@ -30,7 +30,7 @@
 #' knn_model
 #' finalize_model(knn_model, lowest_rmse)
 finalize_model <- function(x, parameters) {
-  if (!inherits(x, "model_spec")) {
+  if (!inherits(x, "model_spec") && !inherits(x, "cluster_spec")) {
     cli::cli_abort(
       "{.arg x} should be a parsnip model specification, not
```
hfrick: Does this error message need updating? i.e., is there something more accurate we can say than "parsnip model specification"?

R/loop_over_all_stages-helpers.R
```diff
@@ -3,6 +3,9 @@
 
 # Note: in loop(), we add more elements for the outcome name(s), and the
 # data partitions
+#' @export
+#' @keywords internal
+#' @rdname empty_ellipses
 make_static <- function(
```
hfrick: We export some of our internal helpers for other tidymodels packages already and have a leading `.` in the name for them. Would there be a reason not to do this here too?

R/parallel.R
```diff
@@ -394,22 +394,25 @@ par_fns <- function(framework) {
   res
 }
 
+#'@export
```
hfrick, line 397: Same question about adding a leading `.` to the name

R/schedule.R
```diff
@@ -151,15 +159,37 @@ get_param_info <- function(wflow) {
...
+  } else {
+    param_info$has_submodel[is.na(param_info$has_submodel)] <- FALSE
+  }
 
   param_info
 }
+
+# Determine the source type for model parameters
+# Returns "model_spec" for parsnip models, "cluster_spec" for tidyclust models
+#' @export
```
hfrick: Same question about adding a leading `.` to the name

R/loop_over_all_stages.R
```diff
@@ -44,7 +47,9 @@
   # Now that we have data, determine the names of the outcome data. NOTE that
   # if an inline function is used (e.g. add_formula(log(mpg) ~ .)), We will
   # potentially change it later. See #1024
-  static$y_name <- outcome_names(static$wflow, data = split$data)
+  if (!is_model_metrics) {
+    static$y_name <- outcome_names(static$wflow, data = split$data)
```
hfrick: Could we have a `outcome_names()` method for tidyclust workflows to just return `NULL`? That should make a few things cleaner.

R/loop_over_all_stages.R
```diff
@@ -478,6 +568,9 @@
   return_tbl
 }
 
+#'@export
```
hfrick: This should have a leading `.` in the name

R/loop_over_all_stages.R
```diff
@@ -144,6 +152,70 @@
...
+          location = location,
+          notes = notes
+        )
+
+        if (!is_failure(current_metrics)) {
+          current_metrics <- dplyr::bind_cols(current_grid, current_metrics)
+          metrics_reserve <- dplyr::bind_rows(metrics_reserve, current_metrics)
+        }
+
+        # Extractions for model-based metrics
```
hfrick, line 173: This looks a lot like the extraction code for prediction models which already has a `TODO modularize this` code comment. I would say that now is the time to look into that rather than adding more code duplication :)

R/schedule.R
```diff
@@ -151,15 +159,37 @@ get_param_info <- function(wflow) {
   model_type <- class(model_spec)[1]
   model_eng <- model_spec$engine
 
-  model_param <- parsnip::get_from_env(paste0(model_type, "_args")) |>
-    dplyr::filter(engine == model_spec$engine) |>
-    dplyr::select(name = parsnip, has_submodel)
+  model_param <- parsnip::get_from_env(paste0(model_type, "_args"))
```
hfrick, line 164: I would comment here that this is NULL for tidyclust workflows, then we can deduct that the `if(!is.null(model_param))` block is for parsnip models

R/schedule.R
```diff
@@ -151,15 +159,37 @@ get_param_info <- function(wflow) {
...
+
+  # Handle models not in parsnip's registry (e.g., cluster_spec from tidyclust)
```
hfrick: I find this comment somewhat confusing since it only applies to the `if` condition, not the contents of the block. Can we replace it with the comment I suggested for a line further up?

R/schedule.R
```diff
@@ -151,15 +159,37 @@ get_param_info <- function(wflow) {
...
+    param_info <- dplyr::left_join(param_info, model_param, by = "name")
+  }
 
-  param_info <- dplyr::left_join(param_info, model_param, by = "name")
-  # Parameters for model engines, preprocessors, and postprocessors will have
-  # NA values for `has_submodel` after this merge. Since they cannot be
-  # submodels, we'll convert them to FALSE
-  param_info$has_submodel[is.na(param_info$has_submodel)] <- FALSE
+  # Parameters not in the registry or engine-specific ones will have NA for
+  # has_submodel after this merge. They are not submodels so convert to FALSE.
```
hfrick: Can we move this comment into the `else` block and say here instead that tidyclust models never use submodels?

inst/loop.qmd
```diff
@@ -180,7 +203,7 @@ result <- remove_log_notes(result)
 
 ## Efficiency
 
-The nested structure avoids redundant computation:
```
hfrick, line 183: Did you remove this sentence on purpose?

hfrick (commented):
Generally a thumbs-up! These are not as many changes as I initially expected. A couple small comments and one central one about the extract code block.

Would we ever expect tidyclust to use submodels? Given that's related to prediction, I assuming the answer is no but wanted to check in on that. 

If we do think it might change in future, we should make `schedule_predict_stage_i()` more flexible, it currently has a filter with a hard-coded `"model_spec"`.

topepo (approved):
Looks great. Nice job on the documentation. 

I did some work to see if we could compartmentalize the "For model-based metrics: compute metrics directly and handle extractions" block, but I don't think that is feasible.

EmilHvitfeldt:
>Would we ever expect tidyclust to use submodels? Given that's related to prediction, I assuming the answer is no but wanted to check in on that.
 If we do think it might change in future, we should make `schedule_predict_stage_i()` more flexible, it currently has a filter with a hard-coded `"model_spec"`.
 
 At this moment i don't think we will need to support tidyclust submodels. none of the methods we support can use it, and none of the methods I have looked at for expansion so far supports it either. So I reckon that that this is the correct thing to do

R/loop_over_all_stages-helpers.R
EmilHvitfeldt: That seems like a reasonable thing to do. and it shouldn't break anything as these functions has always been internal beforehand

R/loop_over_all_stages.R
EmilHvitfeldt: done!

R/loop_over_all_stages-helpers.R
```diff
@@ -22,15 +25,30 @@ make_static <- function(
   if (!inherits(param_info, "parameters")) {
     cli::cli_abort("{.arg param_info} should be a {.cls parameters} object.")
   }
-  if (!inherits(metrics, "metric_set")) {
-    cli::cli_abort("{.arg metrics} should be a {.cls metric_set} object.")
+  if (!is_valid_metric_set(metrics)) {
+    cli::cli_abort("{.arg metrics} should be a metric set object.")
```
hfrick: With that change about the class inheritance of `cluster_metric_set` we can go back to the original code here

R/loop_over_all_stages.R
```diff
@@ -109,7 +112,10 @@
     }
 
     # Update y_name in case the workflow had an inline function like `log(mpg) ~ .`
-    static$y_name <- outcome_names(current_wflow)
+    # Only for prediction-based metrics (model-based metrics do not have outcomes)
+    if (!is_model_metrics) {
+      static$y_name <- outcome_names(current_wflow)
+    }
```
hfrick: With `outcome_names()` returning `NULL` for clustering workflow, we also shouldn't need this anymore, could you check?

R/loop_over_all_stages.R
```diff
@@ -445,10 +497,14 @@
...
-    .notes = list(notes),
-    outcome_names = static$y_name
+    .notes = list(notes)
   )
 
+  # Add outcome_names only for prediction-based metrics
+  if (!is_model_metrics) {
+    return_tbl$outcome_names <- static$y_name
+  }
+
```
hfrick: and one more place which hopefully simplifies with the change to `outcome_names()`

hfrick (approved):
This looks good!  Thank you for pulling out the extract code! One last detail: could you check if the two simplifications regarding the outcome names work?

EmilHvitfeldt:
thank you Hannah!

PR #1142, opened 2026-03-06
TLC for `loop_over_all_stages()`

hfrick:
This is a PR to give loopy a gentle tug and pull via aligning a pattern and making names a bit more expressive. 

It's best read on a commit-by-commit basis: The first one adds a description of concepts and names in loopy/`loop_over_all_stages()` and the following commit try to smooth out some small rough edges:
- We use `current_wflow` as the object we carry through and update to match the current state of the iteration. I've made `current_grid` to the same, instead of having new objects at new stages.
- `rebind_grid()` didn't quickly enough get me to understand why we were rebinding, so I named it `extend_grid()` to suggest that we are extending the `current_grid` by the parameter from/in the next stage.
- `all_` makes me think of "across all iterations", so I've brushed up the names for objects related to "all submodels in this iteration".
- That move makes it easy to rename `pred_reserve` to `all_predictions`, if we choose to.

### DIFF
--- a/R/loop_over_all_stages-helpers.R
+++ b/R/loop_over_all_stages-helpers.R
@@ -338,7 +338,7 @@ process_prediction_data <- function(wflow_fit, static) {
 # ------------------------------------------------------------------------------
 # Misc functions
 
-rebind_grid <- function(...) {
+extend_grid <- function(...) {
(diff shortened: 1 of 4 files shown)


topepo (commented):
It looks great. 

Can you add a link in the `loop*.R` files to the location of the `loop.qmd` file and its relevance?

PR #1147, opened 2026-03-16
Fit resampled calbration model on correct submodel predictions

### LINKED ISSUE #1144: For submodels, resampled calibration models are not trained on the correct predictions
When tuning a model that has both a submodel parameter (e.g., `penalty` in glmnet) and a postprocessor that requires fitting (e.g., probability calibration via tailor), all calibration models are trained on predictions made with the same (and thus usually wrong) submodel parameter value.

The loop correctly generates assessment-set predictions for all submodel values at once via `multi_predict()` and filters them per iteration. But when fitting the postprocessor, `finalize_fit_post()` calls `workflows::.fit_post()`, which internally calls `augment(workflow_mock, calibration_data)` → `predict()`. This predict call uses the submodel parameter value from the model spec — which is whatever `min_grid()` chose for fitting, not the value for the current predict iteration.

``` r
library(tidymodels)

set.seed(1)
dat <- modeldata::sim_regression(500)

dat_split <- initial_validation_split(dat)
dat_rs <- validation_set(dat_split)

cal_post <- tailor() |>
  adjust_numeric_calibration()
glmn_spec <- linear_reg(penalty = tune(), mixture = 1) |>
  set_engine("glmnet")
wflow <- workflow(outcome ~ ., glmn_spec, cal_post)

# A small grid with well-separated penalty values
grd <- tibble::tibble(penalty = c(0.001, 0.1))

set.seed(1)
tune_res <- tune_grid(
  wflow,
  resamples = dat_rs,
  metrics = metric_set(rmse),
  grid = grd,
  control = control_grid(
    extract = function(wf) {
      # Extract the fitted calibration model for inspection
      extract_postprocessor(wf, estimated = TRUE)
    }
  )
)
#> Registered S3 method overwritten by 'butcher':
#>   method                 from    
#>   as.character.dev_topic generics

extracts <- collect_extracts(tune_res)

cal_coef_1 <- extracts$.extracts[[1]]$adjustments[[1]]$results$fit$estimates[[1]]$estimate |>
  coefficients()
cal_coef_2 <- extracts$.extracts[[2]]$adjustments[[1]]$results$fit$estimates[[1]]$estimate |>
  coefficients()

identical(cal_coef_1, cal_coef_2)
#> [1] TRUE
```

<sup>Created on 2026-03-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

hfrick:
closes #1144

Instead of letting `.fit_post()` generate calibration predictions, we now pre-compute them ourselves using the same `predict_all_types()` machinery that handles assessment predictions. For submodels, this predicts all values at once via `multi_predict()`, then filters per iteration — mirroring how assessment predictions already work.

### DIFF
--- a/R/loop_over_all_stages-helpers.R
+++ b/R/loop_over_all_stages-helpers.R
@@ -233,12 +233,14 @@ fit_post_from_predictions <- function(wflow, cal_predictions) {
 
 predict_all_types <- function(
   wflow_fit,
-  processed_data_pred,
+  processed_data,
   static,
-  submodel_grid = NULL
+  submodel_grid = NULL,
+  source = c("pred", "cal")
 ) {
-  .data <- static$data$pred$data
-  .ind <- static$data$pred$ind
+  source <- rlang::arg_match(source)
+  .data <- static$data[[source]]$data
+  .ind <- static$data[[source]]$ind
 
   model_fit <- wflow_fit |> hardhat::extract_fit_parsnip()
 
@@ -253,7 +255,7 @@ predict_all_types <- function(
   for (type_iter in static$pred_types) {
     tmp_res <- predict_wrapper(
       model = model_fit,
-      new_data = processed_data_pred$predictors,
+      new_data = processed_data$predictors,
       type = type_iter,
       eval_time = static$eval_time,
       subgrid = submodel_grid
@@ -289,7 +291,7 @@ predict_all_types <- function(
   }
 
   pred <- pred |>
-    dplyr::full_join(processed_data_pred$outcomes, by = ".row")
+    dplyr::full_join(processed_data$outcomes, by = ".row")
 
   # Add implicitly grouped metric data, if applicable
   metrics_by <- get_metrics_by(static$metrics)
--- a/R/loop_over_all_stages-helpers.R
+++ b/R/loop_over_all_stages-helpers.R
@@ -193,8 +193,15 @@ finalize_fit_post <- function(
   wflow_current,
   data_calibration = NULL,
   grid = NULL,
-  cal_predictions = NULL
+  predictions_calibration = NULL
 ) {
+  if (!is.null(data_calibration) && !is.null(predictions_calibration)) {
+    cli::cli_warn(
+      "Both {.arg data_calibration} and {.arg predictions_calibration} were
+      supplied; using {.arg predictions_calibration}."
+    )
+  }
+
   if (is.null(grid)) {
     grid <- dplyr::tibble()
   }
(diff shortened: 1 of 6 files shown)


R/loop_over_all_stages-helpers.R
```diff
@@ -198,19 +202,44 @@ finalize_fit_post <- function(wflow_current, data_calibration, grid = NULL) {
     finalize_tailor(grid)
   wflow_current <- set_workflow_tailor(wflow_current, post_obj)
 
-  workflows::.fit_post(wflow_current, data_calibration)
+  fit_post_from_predictions(wflow_current, predictions_calibration)
+}
+
+# This mimics `.fit_post()`, except it takes predictions instead of the
```
hfrick: I'm not loving that we now break up the pattern of calling `.fit_<workflow_stage>()` but I think it's valuable here. I'm taking this as an indication that maybe we should change workflows to reflect that.

R/loop_over_all_stages.R
```diff
@@ -102,10 +123,6 @@
 
     grid_with_pre <- current_grid
 
-    if (is_failure(pred_data)) {
-      next
-    }
-
```
hfrick, line 108: This is still around, I only moved it up to line 89 so that we "fail early", i.e., go to the next iteration of the loop immediately if the calculation of `pred_data` fails.

R/loop_over_all_stages.R
```diff
@@ -209,6 +251,38 @@
         has_post <- has_tailor(current_wflow)
         num_iterations_post <- max(nrow(current_sched_pred$post_stage[[1]]), 1)
 
+        # if the postprocessor does not require a fit,
+        # this does not cause data leakage
+        current_cal_pred <- current_pred
```
hfrick, line 256:
```suggestion
        # if the postprocessor does not require a fit,
        # this does not cause data leakage
        current_cal_pred <- current_pred
        
```

Adding an empty line to make it more obvious that the comment only applies to line 256

R/loop_over_all_stages-helpers.R
```diff
@@ -189,7 +189,11 @@ has_tailor_estimated <- function(x) {
 # ------------------------------------------------------------------------------
 # Prediction and postprocessing
 
-finalize_fit_post <- function(wflow_current, data_calibration, grid = NULL) {
+finalize_fit_post <- function(
+  wflow_current,
+  predictions_calibration,
```
EmilHvitfeldt, line 194: I like the name change here

R/loop_over_all_stages-helpers.R
```diff
@@ -198,19 +202,44 @@ finalize_fit_post <- function(wflow_current, data_calibration, grid = NULL) {
...
+
+  post_fit <- tailor::fit(
+    object = tailor_obj,
+    .data = predictions_calibration,
+    outcome = outcome_names,
+    estimate = tidyselect::any_of(c(".pred", ".pred_class")),
+    probabilities = c(
+      tidyselect::contains(".pred_"),
+      -tidyselect::matches("^\\.pred$|^\\.pred_class$")
+    )
```
EmilHvitfeldt, line 224: Maybe this isn't the PR where we do this, but I'm starting to think we need some pred name helpers. Because I'm pretty sure that this selects for `.pred_time`

EmilHvitfeldt (approved):

R/loop_over_all_stages-helpers.R
hfrick: I agree. This is taken from workflows, so that'd be the better place to put that. We also have not yet implemented postprocessing for censored regression models.

R/loop_over_all_stages-helpers.R
hfrick: `data_calibration` is the argument of the workflows machinery so keeping it would have been pretty confusing 😄

R/loop_over_all_stages-helpers.R
topepo:
```suggestion
# This mimics `workflows::.fit_post()`, except it takes predictions instead of the
```

R/loop_over_all_stages-helpers.R
topepo: https://github.com/tidymodels/parsnip/issues/1353

topepo (commented):
Looks great.

Issue #1161, opened 2026-03-17
Deprecate  `val_class_or_null()` and `val_class_and_single()`


hfrick:
## Summary

`val_class_or_null()` and `val_class_and_single()` are exported but marked with TODO comments indicating they should be removed once finetune is updated. Each has a corresponding unexported helper (`check_class_or_null()`, `check_class_and_single()`).

## Locations

`val_class_or_null()` and `check_class_or_null()`:
https://github.com/tidymodels/tune/blob/35448169e0f56fa6951845bf3a4128286c8b8654/R/checks.R#L563-L585

`val_class_and_single()` and `check_class_and_single()`:
https://github.com/tidymodels/tune/blob/35448169e0f56fa6951845bf3a4128286c8b8654/R/checks.R#L587-L609

## Current status

Both `val_class_*` functions are exported for finetune's benefit. The TODO comments on lines 585 and 609 say:

```r
# TODO remove this once finetune is updated
```

Neither function is called within tune itself (R/, tests/, or inst/).

## Suggested fix

Check whether finetune still uses these functions. If so, open an issue on finetune. Deprecate all four functions (`val_class_or_null()`, `check_class_or_null()`, `val_class_and_single()`, `check_class_and_single()`) and remove their export/documentation entries.

hfrick:
prep: https://github.com/tidymodels/finetune/pull/136

PR #1168, opened 2026-03-18
deprecate `finalize_workflow_preprocessor()`

### LINKED ISSUE #1157: Dead exported function: `finalize_workflow_preprocessor()`
## Summary

`finalize_workflow_preprocessor()` is exported (via `@rdname tune-internal-functions`) but has zero callers in R/, tests/, or inst/.

## Location

https://github.com/tidymodels/tune/blob/35448169e0f56fa6951845bf3a4128286c8b8654/R/grid_helpers.R#L115-L129

## What it does

Finalizes a workflow's recipe preprocessor by merging tuning parameter values from a grid into the recipe. If the grid has no columns, it returns the workflow unchanged.

## Why it's dead

The current loop architecture uses `finalize_fit_pre()` (in `R/loop_over_all_stages-helpers.R`) instead, which handles preprocessor finalization as part of the broader fit pipeline.

## Suggested fix

Check whether any downstream packages (e.g., finetune, workflowsets) call this function. If not, deprecate it.

hfrick:
closes #1157

This means that tidyclust and agua will start sending deprecation warnings, more details in the issue comment

Note that adding lifecycle to the description moves the check on the number of dependencies from INFO to NOTE
```
❯ checking package dependencies ... NOTE
```
but doesn't tune have an exception for this? We run into this again in https://github.com/tidymodels/tune/pull/1178 and will also for https://github.com/tidymodels/tune/issues/1161

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,6 +26,7 @@ Imports:
     glue (>= 1.6.2),
     GauPro (>= 0.2.15),
     hardhat (>= 1.4.2),
+    lifecycle,
     parallel,
     parsnip (>= 1.4.1.9002),
     purrr (>= 1.0.0),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -10,8 +10,8 @@ Authors@R: c(
 Description: The ability to tune models is important. 'tune' contains
     functions and classes to be used in conjunction with other
     'tidymodels' packages for finding reasonable values of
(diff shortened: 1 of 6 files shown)


EmilHvitfeldt (approved):
I'm good with this, hoping to do a tidyclust release fairly closely coupled with the tune release anyways

topepo (approved):

PR #1176, opened 2026-03-20
Bring back check on number of rows after preprocessing prediction data

### LINKED ISSUE #1158: Revisit code comment from melodie switch
Do we still need this code snippet and it should be integrated? Or can we delete it?

https://github.com/tidymodels/tune/blob/35448169e0f56fa6951845bf3a4128286c8b8654/R/grid_helpers.R#L1-L25

hfrick:
closes #1158

Byproducts:
- https://github.com/tidymodels/parsnip/issues/1355
- https://github.com/tidymodels/workflows/issues/315


R/loop_over_all_stages-helpers.R
```diff
@@ -345,6 +343,22 @@ process_prediction_data <- function(
   .ind <- static$data[[source]]$ind
 
   processed_data <- forge_from_workflow(.data, wflow_fit)
+
+  if (nrow(processed_data$predictors) != nrow(.data)) {
+    set_name <- if (source == "pred") "assessment" else "calibration"
```
topepo:
```suggestion
  if (source == "pred") {
    set_name <- "assessment" 
  } else {
    set_name <- "calibration"
  }
```

R/loop_over_all_stages-helpers.R
```diff
@@ -345,6 +343,22 @@ process_prediction_data <- function(
   .ind <- static$data[[source]]$ind
 
   processed_data <- forge_from_workflow(.data, wflow_fit)
+
+  if (nrow(processed_data$predictors) != nrow(.data)) {
```
topepo, line 374: I don't recall what this is checking. Do you know? Isn't it possible that the assessment and calibration sets are the same size?

hfrick: This checks that applying the preprocessing doesn't remove rows. It used to check the assessment set only but now we also preprocess the calibration data early, that's why you see the correct dataset name spliced into the error below. The check is always data before forging vs after, not assessment vs calibration.

PR #1178, opened 2026-03-20
Deprecate `check_metrics()`

### LINKED ISSUE #1124: Remove overlap of check_metrics() and check_metrics_arg()
There is some duplicated logic that should be modularized. I'm not even sure if we need both. 

`check_metrics()`: 

https://github.com/tidymodels/tune/blob/fc5711b2ae875751e0e7b9df45583cffc5a00708/R/checks.R#L391-L408

`check_metrics_arg()`: 

https://github.com/tidymodels/tune/blob/fc5711b2ae875751e0e7b9df45583cffc5a00708/R/metric-selection.R#L311-L325

### LINKED ISSUE #1160: Deprecate superseded `check_metrics()`
## Summary

`check_metrics()` is exported and documented (via `@rdname empty_ellipses`) but is only called from tests. It has been replaced by `check_metrics_arg()` in the production code paths.

## Location

https://github.com/tidymodels/tune/blob/35448169e0f56fa6951845bf3a4128286c8b8654/R/checks.R#L384-L456

## What it does

`check_metrics()` provides default metric sets based on model mode (regression, classification, censored regression, quantile regression) and validates that user-supplied metrics match the model mode.

## Why it's superseded

The current code uses `check_metrics_arg()` instead. `check_metrics()` has no callers in R/ or inst/ -- it is only exercised by tests that were written against it directly.

## Suggested fix

Since it is exported and documented, check whether any downstream packages (e.g., finetune, workflowsets) call it. Set the deprecation process for `check_metrics()` accordingly.

hfrick:
Closes #1160 and closes #1124

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,6 +26,7 @@ Imports:
     glue (>= 1.6.2),
     GauPro (>= 0.2.15),
     hardhat (>= 1.4.2),
+    lifecycle,
     parallel,
     parsnip (>= 1.4.1.9002),
     purrr (>= 1.0.0),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -10,8 +10,8 @@ Authors@R: c(
 Description: The ability to tune models is important. 'tune' contains
     functions and classes to be used in conjunction with other
     'tidymodels' packages for finding reasonable values of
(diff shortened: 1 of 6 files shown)


topepo (approved):
🥳  This brings me joy

PR #1191, opened 2026-05-18
unlist tuning parameters wrapped in lists 

### LINKED ISSUE #1190: avoid failures with tune parameters that are not scalars
We are coming up on models that can have non-scale values, list as vector or other object types. Right now, if they are in list columns of a grid, the stay lists in the splined object. 

Here's an example for ResNet, which can have integer vector arguments: 

```
tab resnet Model Specification (classification)

Main Arguments:
  hidden_units = list(c(2L, 4L, 9L))
  batch_norm_units = list(c(12L, 19L, 24L))
  penalty = 1.83298071083243e-05
  epochs = 50
  activation = relu
  learn_rate = 0.119935394620923

Engine-Specific Arguments:
  stop_iter = 5
  residual_at = 2:3
  optimizer = ADAMw
  verbose = FALSE
  rate_schedule = step
  batch_size = 114
  momentum = 0.99

Computational engine: brulee 
```

topepo:
Closes #1190

### DIFF
--- a/R/finalize.R
+++ b/R/finalize.R
@@ -51,6 +51,10 @@ finalize_model <- function(x, parameters) {
         discordant$name[i]
     }
   }
+
+  # In case the parameter is not a scalar value
+  parameters <- delist(parameters)
+
   rlang::exec(update, object = x, !!!parameters)
 }
 
@@ -68,6 +72,9 @@ finalize_recipe <- function(x, parameters) {
   if (tibble::is_tibble(parameters)) {
     parameters <- as.list(parameters)
   }
+
+  # In case the parameter is not a scalar value
+  parameters <- delist(parameters)
 
   parameters <- parameters[names(parameters) %in% pset$id]
   parameters <- parameters[pset$id]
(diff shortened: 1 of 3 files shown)


R/finalize.R
```diff
@@ -170,3 +180,14 @@ complete_steps <- function(param, pset, object) {
...
 }
+
+delist <- function(x) {
+  grid <- as.list(x)
+  is_list_param <- purrr::map_lgl(grid, is.list)
+  if (any(is_list_param)) {
+    for (inx in which(is_list_param)) {
+      grid[[inx]] <- grid[[inx]][[1]]
+    }
+  }
```
EmilHvitfeldt, line 191: not a big deal, but like this better

```suggestion
  if (!any(is_list_param)) {
    return(grid)
  }
  
  for (inx in which(is_list_param)) {
    grid[[inx]] <- grid[[inx]][[1]]
  }
```

EmilHvitfeldt (approved):
very nice, small PR

